PR #21, opened 2021-05-29
Rph/like-reclout-diamond-endpoints

redpartyhat:


routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+		return
+	}
+
+	// Filter out any restricted profiles.
+	addReaderPublicKey := false
+	pkMapToFilter := make(map[lib.PkMapKey][]byte)
+	for _, pubKey := range reclouterPubKeys {
+		pkMapKey := lib.MakePkMapKey(pubKey)
+		pkMapToFilter[pkMapKey] = pubKey
+		if reflect.DeepEqual(pubKey, readerPublicKeyBytes) {
```
lazynina, line 1687: can we skip this reflect.DeepEqual call if we have `addReaderPublicKey = true` already? `reflect.DeepEqual` is relatively expensive so if we can avoid that call, it would help performance. 

Actually it would probably be best if we look up the like entry directly for the reader instead of comparing every public key to the reader.
```
	postEntry := utxoView.GetPostEntryForPostHash(postHash)
	postEntryReaderState := utxoView.GetPostEntryReaderState(readerPublicKeyBytes, postEntry)
	addReaderPublicKey := postEntryReaderState.LikedByReader
```

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+			likers = append(likers, profileEntryResponse)
+		}
+	}
+
+	// Almost done. Just need to sort the likers.
+	sort.Slice(likers, func(ii, jj int) bool {
+
+		// Attempt to sort on bitclout locked.
+		iiBitCloutLocked := likers[ii].CoinEntry.BitCloutLockedNanos
+		jjBitCloutLocked := likers[jj].CoinEntry.BitCloutLockedNanos
```
lazynina, line 1726: How are we handling likes from anonymous users?  this looks like it would raise a nil pointer.  okay actually I see that we skip anonymous public key above -- is this the expected behavior? I think it's fine.

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+		return
+	} else {
+		postHashBytes, err := hex.DecodeString(requestData.PostHashHex)
+		if err != nil || len(postHashBytes) != lib.HashSizeBytes {
+			_AddBadRequestError(ww, fmt.Sprintf("GetLikesForPost: Error parsing post hash %v: %v",
+				requestData.PostHashHex, err))
+			return
+		}
+		postHash = &lib.BlockHash{}
+		copy(postHash[:], postHashBytes)
```
lazynina, line 1652: nit: this doesn't need to be in an else statement

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+		return
+	} else {
+		postHashBytes, err := hex.DecodeString(requestData.PostHashHex)
+		if err != nil || len(postHashBytes) != lib.HashSizeBytes {
+			_AddBadRequestError(ww, fmt.Sprintf("GetDiamondsForPost: Error parsing post hash %v: %v",
+				requestData.PostHashHex, err))
+			return
+		}
+		postHash = &lib.BlockHash{}
+		copy(postHash[:], postHashBytes)
```
lazynina, line 1794: this doesn't need to be in an else statement

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+	}
+
+	// Decode the reader public key into bytes. Default to nil if no pub key is passed in.
+	var readerPublicKeyBytes []byte
+	if requestData.ReaderPublicKeyBase58Check != "" {
+		var err error
+		readerPublicKeyBytes, _, err = lib.Base58CheckDecode(requestData.ReaderPublicKeyBase58Check)
+		if requestData.ReaderPublicKeyBase58Check != "" && err != nil {
```
lazynina, line 1802: don't need to check that `requestData.ReaderPublicKeyBase58Check != ""`

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+			pkMapToFilter[pkMapKey] = profileEntry.PublicKey
+		}
+	}
+	filteredPkMap, err := fes.FilterOutRestrictedPubKeysFromMap(pkMapToFilter, readerPublicKeyBytes, "leaderboard" /*moderationType*/)
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("GetDiamondsForPost: Error filtering out restricted profiles: %v", err))
+		return
+	}
+
+	// Create a list of unifltered (aka not blacklisted) diamondSenders.
```
lazynina, line 1838: `unifltered` -> `unfiltered`

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+	diamondSenders := []*lib.ProfileEntry{}
+	for senderPKID := range pkidToDiamondLevel {
+		profileEntry := utxoView.GetProfileEntryForPKID(&senderPKID)
+		if profileEntry == nil {
+			continue
+		} else {
+			if _, ok := filteredPkMap[lib.MakePkMapKey(profileEntry.PublicKey)]; ok {
+				diamondSenders = append(diamondSenders, profileEntry)
```
lazynina, line 1846: doesn't need to be inside an else

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+	})
+
+	// Cut out the page of diamondSenders that we care about.
+	diamondSendersLength := uint32(len(diamondSenders))
+	// Slice the comments from the offset up to either the end of the slice or the offset + limit, whichever is smaller.
+	maxIdx := lib.MinUint32(diamondSendersLength, requestData.Offset+requestData.Limit)
+	var diamondSendersPage []*lib.ProfileEntry
+	if diamondSendersLength > requestData.Offset {
+		diamondSendersPage = diamondSenders[requestData.Offset:maxIdx]
+	}
```
lazynina, line 1885: this appears incorrect to me.  I think we always want to take a subset of the slice from `[requestData.Offset:maxIdx]`, not only when Offset is greater than diamondSendersLength?

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+		return
+	} else {
+		postHashBytes, err := hex.DecodeString(requestData.PostHashHex)
+		if err != nil || len(postHashBytes) != lib.HashSizeBytes {
+			_AddBadRequestError(ww, fmt.Sprintf("GetRecloutsForPost: Error parsing post hash %v: %v",
+				requestData.PostHashHex, err))
+			return
+		}
+		postHash = &lib.BlockHash{}
+		copy(postHash[:], postHashBytes)
```
lazynina, line 1948: does not need to be in an else

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+	}
+
+	// Decode the reader public key into bytes. Default to nil if no pub key is passed in.
+	var readerPublicKeyBytes []byte
+	if requestData.ReaderPublicKeyBase58Check != "" {
+		var err error
+		readerPublicKeyBytes, _, err = lib.Base58CheckDecode(requestData.ReaderPublicKeyBase58Check)
+		if requestData.ReaderPublicKeyBase58Check != "" && err != nil {
```
lazynina, line 1956: don't need to check `requestData.ReaderPublicKeyBase58Check != ""` here

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+
+	// Fetch the reclouters for the post requested.
+	reclouterPubKeys, err := utxoView.GetRecloutsForPostHash(postHash)
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("GetRecloutsForPost: Error getting reclouters %v", err))
+		return
+	}
+
+	// Filter out any restricted profiles.
+	addReaderPublicKey := false
```
lazynina, line 1977: I think we can get `addReaderPublicKey` in a similar fashion to my comment on the `GetLikesForPost` about using `GetPostEntryReaderState` instead of comparing every public key to the reader

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+
+	// Create a list of the reclouters that were not restricted.
+	reclouters := []*ProfileEntryResponse{}
+	for _, filteredPubKey := range filteredPkMap {
+		profileEntry := utxoView.GetProfileEntryForPublicKey(filteredPubKey)
+		if profileEntry == nil {
+			continue
+		} else {
+			profileEntryResponse := _profileEntryToResponse(profileEntry, fes.Params, verifiedMap, utxoView)
+			reclouters = append(reclouters, profileEntryResponse)
```
lazynina, line 2013: nit: this doesn't need to be in an else

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+	})
+
+	// Cut out the page of reclouters that we care about.
+	recloutersLength := uint32(len(reclouters))
+	// Slice the comments from the offset up to either the end of the slice or the offset + limit, whichever is smaller.
+	maxIdx := lib.MinUint32(recloutersLength, requestData.Offset+requestData.Limit)
+	var recloutersPage []*ProfileEntryResponse
+	if recloutersLength > requestData.Offset {
+		recloutersPage = reclouters[requestData.Offset:maxIdx]
+	}
```
lazynina, line 2040: I think we always want to take the slice of `[requestData.Offset:maxIdx]`

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+	} else {
+		postHashBytes, err := hex.DecodeString(requestData.PostHashHex)
+		if err != nil || len(postHashBytes) != lib.HashSizeBytes {
+			_AddBadRequestError(ww, fmt.Sprintf(
+				"GetQuoteRecloutsForPost: Error parsing post hash %v: %v",
+				requestData.PostHashHex, err))
+			return
+		}
+		postHash = &lib.BlockHash{}
+		copy(postHash[:], postHashBytes)
```
lazynina, line 2087: nit: this doesn't need to be in an else clause

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+	})
+
+	// Cut out the page of reclouters that we care about.
+	quoteRecloutsLength := uint32(len(quoteReclouts))
+	// Slice the comments from the offset up to either the end of the slice or the offset + limit, whichever is smaller.
+	maxIdx := lib.MinUint32(quoteRecloutsLength, requestData.Offset+requestData.Limit)
+	var quoteRecloutsPage []*PostEntryResponse
+	if quoteRecloutsLength > requestData.Offset {
+		quoteRecloutsPage = quoteReclouts[requestData.Offset:maxIdx]
+	}
```
lazynina, line 2189: I think we always want to take the slice from `[requestData.Offset:maxIdx]`

routes/server.go
```diff
@@ -791,6 +796,34 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
 			fes.GetSinglePost,
 			false,
 		},
+		{
+			"BlockGetTxn",
```
lazynina, line 800: I think this should be `GetLikesforPost`

routes/server.go
```diff
@@ -791,6 +796,34 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
...
+			[]string{"POST", "OPTIONS"},
+			RoutePathGetLikesForPost,
+			fes.GetLikesForPost,
+			false,
+		},
+		{
+			"BlockGetTxn",
```
lazynina, line 807: `GetDiamondsForPost`

routes/server.go
```diff
@@ -791,6 +796,34 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
...
+			[]string{"POST", "OPTIONS"},
+			RoutePathGetDiamondsForPost,
+			fes.GetDiamondsForPost,
+			false,
+		},
+		{
+			"BlockGetTxn",
```
lazynina, line 814: `GetRecloutsForPost`

routes/server.go
```diff
@@ -791,6 +796,34 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
...
+			[]string{"POST", "OPTIONS"},
+			RoutePathGetRecloutsForPost,
+			fes.GetRecloutsForPost,
+			false,
+		},
+		{
+			"BlockGetTxn",
```
lazynina, line 821: `GetQuoteRecloutsForPost`

lazynina (changes requested):

routes/post.go
```diff
@@ -1599,17 +1600,601 @@ func (fes *APIServer) GetDiamondedPosts(ww http.ResponseWriter, req *http.Reques
...
+		// Sort based on pub key if all else fails.
+		return likers[ii].PublicKeyBase58Check > likers[jj].PublicKeyBase58Check
+	})
+
+	// Cut out the page of reclouters that we care about.
+	likersLength := uint32(len(likers))
+	// Slice the comments from the offset up to either the end of the slice or the offset + limit, whichever is smaller.
+	maxIdx := lib.MinUint32(likersLength, requestData.Offset+requestData.Limit)
+	var likersPage []*ProfileEntryResponse
+	if likersLength > requestData.Offset {
```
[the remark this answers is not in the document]
lazynina: Okay so I see why you have the if statement written this way, but I would expect to get a slice of length 0 in the response instead of nil. Can we initialize likersPage to an empty slice? this applies to all the `Get{INTERACTION}ForPost` functions

lazynina (approved):
:+1:

PR #28, opened 2021-06-01
only report once a minute to datadog, return true for IsConfigureForWyre if the WyreURL is set, ignore all other flags

lazynina:


maebeam (approved):

maebeam:
merge away!

PR #36, opened 2021-06-08
Super Admin Public Keys

lazynina:
add super admin public keys flag and separate out super admin routes. move admin public keys flag to this repo

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -31,6 +31,7 @@ type Config struct {
 	SecureHeaderDevelopment   bool
 	SecureHeaderAllowHosts    []string
 	AdminPublicKeys           []string
+	SuperAdminPublicKeys      []string
 
 	// Analytics + Profiling
 	AmplitudeKey           string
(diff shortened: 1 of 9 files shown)


maebeam:
Would you mind cleaning this code up a little bit by adding a new type which we can use in the `Route` struct and as a param to `CheckAdmin`? I'm thinking something like:

```
type AccessLevel int

const (
    NoAccess AccessLevel = iota
    AdminAccess
    SuperAdminAccess
)
```

lazynina:
> Would you mind cleaning this code up a little bit by adding a new type which we can use in the `Route` struct and as a param to `CheckAdmin`? I'm thinking something like:
> 
> ```
> type AccessLevel int
> 
> const (
>     PublicAccess AccessLevel = iota
>     AdminAccess
>     SuperAdminAccess
> )
> ```

make sense to me!

maebeam (approved):
looks great!

PR #40, opened 2021-06-10
Get single profile picture endpoint

lazynina:
add get single profile picture endpoint, remove profile pic from profile entry response to lighten payloads

### DIFF
--- a/routes/user.go
+++ b/routes/user.go
@@ -162,9 +162,9 @@ func (fes *APIServer) updateUserFieldsStateless(user *User, utxoView *lib.UtxoVi
 	user.PublicKeysBase58CheckFollowedByUser = publicKeysBase58CheckFollowedByUser
 	pkid := utxoView.GetPKIDForPublicKey(publicKeyBytes)
 	if !skipHodlings {
-		// Get the users that the user hodls and vice versa
-		youHodlMap, hodlYouMap, err := fes.GetHodlingsForPublicKey(
-			pkid, true /*fetchProfiles*/, utxoView)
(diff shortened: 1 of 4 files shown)


routes/user.go
```diff
@@ -928,6 +916,55 @@ func (fes *APIServer) augmentProfileEntry(
...
+	}
+	publicKeyBytes, _, err := lib.Base58CheckDecode(publicKeyBase58Check)
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("GetSingleProfilePicture: Problem decoding user public key: %v", err))
+		return
+	}
+	// We ignore errors here and simply return
+	profilePicture, _ := fes._getProfilePictureForPublicKey(publicKeyBytes)
+
+	if err = json.NewEncoder(ww).Encode(GetSingleProfilePictureResponse{
```
maebeam, line 960: We should just serve the image itself as bytes so browsers recognize the MIME type and we get free client and edge caching

maebeam: If no image, return a 404 and let the frontend use the default_profile_pic

maebeam (approved):

harpreet9629:
How we can get single profile picture with api?

lazynina:
> How we can get single profile picture with api?

GET /api/v0/get-single-profile-picture/{PublicKeyBase58Check}

^thats the endpoint

harpreet9629:
I m getting error on send bitclout
Error: SendBitclout: User with name not found!

I m using public key for both sender and receiver

PR #59, opened 2021-06-19
Track last hour of pricing data

lazynina:
keep one hours worth of last trade price history and take the max of those values

### DIFF
--- a/routes/base.go
+++ b/routes/base.go
@@ -181,20 +181,23 @@ func (fes *APIServer) UpdateUSDCentsToBitCloutExchangeRate() {
 // from valid elements.
 func (fes *APIServer) getMaxPriceFromHistoryAndCull(currentTimestamp uint64) uint64 {
 	maxPrice := uint64(0)
-	i := 0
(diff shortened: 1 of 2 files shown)


routes/base.go
```diff
@@ -154,18 +155,49 @@ func (fes *APIServer) UpdateUSDCentsToBitCloutExchangeRate() {
...
 
 	glog.Infof("Final exchange rate: %v", fes.UsdCentsPerBitCloutExchangeRate)
 }
 
+// getMaxPriceFromHistoryAndCull removes elements that are outside of the lookback window and return the max price
+// from valid elements.
+func (fes *APIServer) getMaxPriceFromHistoryAndCull(currentTimestamp uint64) uint64 {
+	maxPrice := uint64(0)
+	i := 0
+	for _, priceHistoryItem := range fes.LastTradeBitCloutPriceHistory {
```
diamondhands0, line 185: Any reason we didn't do:
```
for ii, PriceHistoryIdem := range fes.LastTradeBitCloutPriceHistory
```

PR #62, opened 2021-06-28
Ln/txn ids only

lazynina:
The documentation described for [look up transactions for a public key](https://docs.bitclout.com/devs/exchange-listing-api#look-up-transactions-for-a-public-key) and for the [transaction-info endpoint](https://docs.bitclout.com/devs/exchange-listing-api#api-v-1-transaction-info) explain an `IDsOnly` boolean that can be used to only return the TransactionBase58Check attribute of a transaction.

This resolves @carsenk's issue described [here](https://github.com/bitclout/core/issues/46#issuecomment-868699768)

### DIFF
--- a/routes/exchange.go
+++ b/routes/exchange.go
@@ -735,8 +735,6 @@ type APITransactionInfoResponse struct {
 	// to newest.
 	Transactions []*TransactionResponse
 
-	TransactionIDs []string
-
 	BalanceNanos uint64
 }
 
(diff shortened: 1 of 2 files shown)


maebeam (approved):

carsenk:
@lazynina @maebeam this seems to return an empty array now with IDsOnly: true

image

lazynina:
> @lazynina @maebeam this seems to return an empty array now with IDsOnly: true
> 
> image

I can't reproduce this behavior. It seems like your TxIndex isn't built.

carsenk:
@lazynina You are correct, my node isnt synced, seems to be stuck 1606 blocks out, I will redo the node

PR #69, opened 2021-07-01
Add IsFollowingPublicKey endpoint

mubashariqbal:
This is a convenience endpoint to check if one user is following another.

I have a number of projects that need to know if one user is following another, and it currently requires downloading all the follow entries to be able to check.

This endpoint would remove that need.


jacklalley:
Nice @mubashariqbal. 

I think it would be good for the core dev team to set some precedent on this. 

Is the BitClout API intended to be a general-purpose API for developers building on the blockchain itself or is it intended to be specifically built for just the bitclout.com platform? 

If the latter then it may be worth the community branching off the API via a node so that it can be extended with endpoints useful for 3rd party applications.

routes/user.go
```diff
@@ -2331,3 +2331,79 @@ func (fes *APIServer) BlockPublicKey(ww http.ResponseWriter, req *http.Request)
...
+	followEntries, err = utxoView.GetFollowEntriesForPublicKey(userPublicKeyBytes, false)
+
+	for _, followEntry := range followEntries {
+		// get the profile entry for each follower pubkey
+		var followPKID *lib.PKID
+		followPKID = followEntry.FollowedPKID
+
+		// Convert the followPKID to a public key using the view. The followPubKey should never
+		// be nil.
+		followPubKey := utxoView.GetPublicKeyForPKID(followPKID)
```
lazynina, line 2392: correct that this should never be nil, but we should check that the length is non-zero just to be safe.
```suggestion
		followPubKey := utxoView.GetPublicKeyForPKID(followPKID)
		if len(followPubKey) == 0 {
			_AddBadRequestError(ww, fmt.Sprintf("IsFollowingPublicKey: found PKID %v that does not have a public key mapping; this should never happen", lib.PkToString(followEntry.FollowedPKID[:], fes.Params))
		}
```

routes/user.go
```diff
@@ -2331,3 +2331,79 @@ func (fes *APIServer) BlockPublicKey(ww http.ResponseWriter, req *http.Request)
...
+	utxoView, err := fes.backendServer.GetMempool().GetAugmentedUniversalView()
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("IsFollowingPublicKey Error getting view: %v", err))
+		return
+	}
+
+	var isFollowing = false
+
+	followEntries := []*lib.FollowEntry{}
+	followEntries, err = utxoView.GetFollowEntriesForPublicKey(userPublicKeyBytes, false)
```
lazynina, line 2383: Handle this error

routes/user.go
```diff
@@ -2331,3 +2331,79 @@ func (fes *APIServer) BlockPublicKey(ww http.ResponseWriter, req *http.Request)
...
+
+	for _, followEntry := range followEntries {
+		// get the profile entry for each follower pubkey
```
lazynina, line 2386: we need to check somewhere in here that `followEntry.isDeleted` is false. 
```suggestion
		if followEntry.isDeleted {
			continue
		}
		// get the profile entry for each follower pubkey
```

lazynina (changes requested):
This looks pretty good -- a few small requested changes.

mubashariqbal:
@lazynina Thanks for the review, will update the PR later today.

routes/user.go
mubashariqbal: @lazynina I tried to add this in my local branch, but I'm getting an error when running. Since isDeleted is not exported is is not accessible from the backend codebase.

```
routes/user.go:2390:17: followEntry.isDeleted undefined (cannot refer to unexported field or method isDeleted)
```

lazynina: that's right - apologies, disregard that comment then please.

routes/user.go
mubashariqbal: Added check below..

routes/user.go
mubashariqbal: Added check per suggeston.

mubashariqbal:
Requested changes have been made.

lazynina (approved):
Great work @mubashariqbal!

PR #70, opened 2021-07-04
Add shared secret encryption/decryption

AeonSw4n:
Backend receives EncryptedMessageText (hex) encrypted using shared secret from the frontend. V2 to distinguish between new and legacy encryption schemes.

### DIFF
--- a/routes/message.go
+++ b/routes/message.go
@@ -299,6 +299,11 @@ func (fes *APIServer) getMessagesStateless(publicKeyBytes []byte,
 			}
 		}
 
+		V2 := false
+		if messageEntry.Version == 2{
+			V2 = true
+		}
+
 		// By now we know this messageEntry is meant to be included in the response.
 		messageEntryRes := &MessageEntryResponse{
 			SenderPublicKeyBase58Check:    lib.PkToString(messageEntry.SenderPublicKey, fes.Params),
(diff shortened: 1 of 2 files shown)


routes/message.go
```diff
@@ -439,7 +440,7 @@ func (fes *APIServer) GetMessagesStateless(ww http.ResponseWriter, rr *http.Requ
 type SendMessageStatelessRequest struct {
 	SenderPublicKeyBase58Check    string `safeForLogging:"true"`
 	RecipientPublicKeyBase58Check string `safeForLogging:"true"`
-	MessageText                   string
+	EncryptedMessageText          string
```
maebeam, line 443: I'd like to maintain backwards compatibility for at least two weeks to give apps enough time to update. Let's keep `MessageText` and add `EncryptedMessageText`

routes/message.go
```diff
@@ -483,7 +484,7 @@ func (fes *APIServer) SendMessageStateless(ww http.ResponseWriter, req *http.Req
 	// Try and create the message for the user.
 	tstamp := uint64(time.Now().UnixNano())
 	txn, totalInput, changeAmount, fees, err := fes.blockchain.CreatePrivateMessageTxn(
-		senderPkBytes, recipientPkBytes, requestData.MessageText,
+		senderPkBytes, recipientPkBytes, requestData.EncryptedMessageText,
```
maebeam, line 487: To maintain backwards compatibility we'll need an extra param here. I think one param for `MessageText` and one for `EncryptedMessageText` works fine.

maebeam (approved):

PR #74, opened 2021-07-09
shebang compatible with *nix systems

shrimalmadhur:
Currently when I ran these node scripts on my MacOSX system it wasn't able to resolve the shebang `/bin/env`.
```
zsh: ./n0_test: bad interpreter: /bin/env: no such file or directory
```

Reading more I found out that `/usr/bin/env` is more compatible with most `*nix` based systems. This PR is to change it.

For reference - https://www.cyberciti.biz/tips/finding-bash-perl-python-portably-using-env.html

OS | env path
-- | --
OpenBSD | /usr/bin/env
FreeBSD | /usr/bin/env
Debian | /usr/bin/env
Ubuntu | /usr/bin/env
CentOS | /usr/bin/env
macOS | /usr/bin/env
SUSE | /usr/bin/env
RHEL | /usr/bin/env
NetBSD | /usr/bin/env
Solaris | /usr/bin/env

### DIFF
--- a/scripts/nodes/n0
+++ b/scripts/nodes/n0
@@ -1,4 +1,4 @@
-#!/bin/env bash
+#!/usr/bin/env bash
 
 DATE=`date +%Y%m%d-%H:%M:%S-%N`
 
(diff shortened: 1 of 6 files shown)


maebeam:
Thank you! Been meaning to make this change for a long time

PR #76, opened 2021-07-09
add default limit to comments

shrimalmadhur:
As mentioned in https://github.com/bitclout/backend/issues/67 - if a client doesn't provides a default `CommentLimit` to the endpoint, we don't return any comments.  I think the API should be such that if we have 0 or -ve limit provided by client, we should just use some default value (I'm using 10 here).

### DIFF
--- a/routes/post.go
+++ b/routes/post.go
@@ -1266,7 +1266,7 @@ func (fes *APIServer) GetSinglePost(ww http.ResponseWriter, req *http.Request) {
 	})
 
 	commentEntryResponseLength := uint32(len(commentEntryResponseList))
-	// If a user provide a default or -ve limit, default to 10
+	// If a client provides a default(0) or -ve limit, default to 10
 	if requestData.CommentLimit <= 0 {
 		requestData.CommentLimit = 10
 	} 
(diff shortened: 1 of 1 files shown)


routes/post.go
```diff
@@ -1266,6 +1266,10 @@ func (fes *APIServer) GetSinglePost(ww http.ResponseWriter, req *http.Request) {
 	})
 
 	commentEntryResponseLength := uint32(len(commentEntryResponseList))
+	// If a client provides a default(0) or -ve limit, default to 10
+	if requestData.CommentLimit <= 0 {
```
lazynina, line 1270: `CommentLimit` is a uint32, so it will never be less than 0. Additionally, this prevents the API from only fetching the post without comments. Users should specify CommentLimit on this endpoint to fetch the number of comments they want.

lazynina (commented):
I don't think we want to change this as it prevents the ability of requesting a single post without getting any comments.

routes/post.go
shrimalmadhur: I see. this makes sense!

shrimalmadhur:
@lazynina yea that makes sense. Post without comments might be needed. Didn't think of this. Thanks for explanation. I will close this.

PR #77, opened 2021-07-10
Add IsHodlingPublicKey endpoint

### LINKED ISSUE #75: Add API Endpoint: IsHodling
Add a new endpoint that returns the creator coin holding information of one user by another user

More details: 
https://github.com/bitclout/cips/discussions/4

mubashariqbal:
Fixes #75 

New endpoint to check if one public key is holding the creator coin of another public key, returns true/flag flag and the BalanceEntry if coin is being held.

### DIFF
--- a/routes/user.go
+++ b/routes/user.go
@@ -2413,3 +2413,81 @@ func (fes *APIServer) IsFollowingPublicKey(ww http.ResponseWriter, req *http.Req
 		return
 	}
 }
+
+type IsHodlingPublicKeyRequest struct {
+	PublicKeyBase58Check            string
+	IsHodlingPublicKeyBase58Check	string
+}
+
+type IsHodlingPublicKeyResponse struct {
+	IsHodling 		bool
+	BalanceEntry 	*BalanceEntryResponse	
+}
+
+func (fes *APIServer) IsHodlingPublicKey(ww http.ResponseWriter, req *http.Request) {
+
+	decoder := json.NewDecoder(io.LimitReader(req.Body, MaxRequestBodySizeBytes))
+	requestData := IsHodlingPublicKeyRequest{}
+	if err := decoder.Decode(&requestData); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf(
+			"IsHodlingPublicKey: Problem parsing request body: %v", err))
+		return
+	}
+
+	var userPublicKeyBytes []byte
+	var err error
+	userPublicKeyBytes, _, err = lib.Base58CheckDecode(requestData.PublicKeyBase58Check)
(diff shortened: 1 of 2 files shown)


routes/user.go
```diff
@@ -2413,3 +2413,81 @@ func (fes *APIServer) IsFollowingPublicKey(ww http.ResponseWriter, req *http.Req
...
+		return
+	}
+
+	for _, hodl := range youHodlMap {
+		if hodl.CreatorPublicKeyBase58Check == requestData.IsHodlingPublicKeyBase58Check {
+			BalanceEntry = hodl
+			IsHodling = true
+			break
+		}
+	}
```
lazynina, line 2481: we should get the PKID for the `IsHodlingPublicKeyBase58Check` and just get it from the map.

routes/user.go
```diff
@@ -2413,3 +2413,81 @@ func (fes *APIServer) IsFollowingPublicKey(ww http.ResponseWriter, req *http.Req
...
+
+	res := IsHodlingPublicKeyResponse{
+		IsHodling: 		IsHodling,
```
lazynina, line 2484: fix spacing please.

lazynina (changes requested):
I think this endpoint would be far more efficient if we expose the function `_getBalanceEntryForHODLerPubKeyAndCreatorPubKey` on line 3766 in block_view.go in core.

let's make that function public by changing its name to `GetBalanceEntryForHODLerPubKeyAndCreatorPubKey` and use that to directly get the balance entry instead of getting the whole map.

@maebeam any concerns about making that function public?

routes/user.go
lazynina: this comment becomes irrelevant with my proposal above.

routes/user.go
mubashariqbal: Fixed this and a couple of other issues.

routes/user.go
mubashariqbal: Will wait to hear which route I should take.

maebeam:
> @maebeam any concerns about making that function public?
Make it public

routes/user.go
lazynina: once [this PR](https://github.com/bitclout/core/pull/65) gets merged in, please update to `lib.GetBalanceEntryForHODLerPubKeyAndCreatorPubKey` to directly get the BalanceEntry instead of fetching all the balance entries.

lazynina: @mubashariqbal that function is now public and you should be able to make the appropriate changes.

mubashariqbal: Update the endpoint to use the new function, thanks!

lazynina (approved):
:+1:

lazynina:
this is good to go.

PR #78, opened 2021-07-10
Add Postgres, GraphQL, and Dataloaden

maebeam:


tijno:
@maebeam will global state move to postgres too? im looking at whether its worth creating global state db that can handle states from multiple nodes via `global-state-remote-node`. But if its moving to postgres then its easier to wait for that.

maebeam:
It is moving to postgres (or mysql, exploring that option this week)

tijno:
Ok thx for letting me know @maebeam . My votes for sticking with postgres and not mysql, and maybe having a separate DB url for it to allow optional storing of global state separate from block / deso data.

One request would be to allow some kind of "node" classifier field in global state so it can maintain state for multiple nodes separately in a single DB.

maebeam:
I'm encountering performance issues with postgres so I'm evaluating the performance of mysql. Unfortunately if it is significantly faster we will have to switch

maebeam:
I will keep that request in mind when moving global state to mysql / postgres

PR #83, opened 2021-07-20
Add email verification flow

maebeam:
I also cleaned up the large number of params we were passing to APIServer.

### DIFF
--- a/routes/verify.go
+++ b/routes/verify.go
@@ -6,6 +6,8 @@ import (
 	"encoding/gob"
 	"encoding/json"
 	"fmt"
+	"github.com/sendgrid/sendgrid-go"
+	"github.com/sendgrid/sendgrid-go/helpers/mail"
 	"io"
 	"net/http"
 	"net/url"
@@ -13,6 +15,7 @@ import (
 	"strings"
 
 	"github.com/bitclout/core/lib"
+	"github.com/golang/glog"
 	"github.com/nyaruka/phonenumbers"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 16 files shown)


routes/verify.go
```diff
@@ -318,3 +321,94 @@ func (fes *APIServer) SubmitPhoneNumberVerificationCode(ww http.ResponseWriter,
...
+	validHash := fes.verifyEmailHash(userMetadata.Email, requestData.PublicKey)
+	if requestData.EmailHash != validHash {
+		_AddBadRequestError(ww, fmt.Sprintf("VerifyEmail: Invalid hash: %s", requestData.EmailHash))
+		return
+	}
+
+	userMetadata.EmailVerified = true
+
+	err = fes.putUserMetadataInGlobalState(userMetadata)
+	if err != nil {
```
lazynina, line 360:
```suggestion
	if err := fes.putUserMetadataInGlobalState(userMetadata); err != nil {
```

routes/verify.go
```diff
@@ -318,3 +321,94 @@ func (fes *APIServer) SubmitPhoneNumberVerificationCode(ww http.ResponseWriter,
...
+		_AddBadRequestError(ww, fmt.Sprintf("VerifyEmail: Failed to save user: %v", err))
+		return
+	}
```
lazynina, line 363: Do we need to send back some sort of 200 response at the end of this function? or does it automatically return a 200 if nothing is sent back?

lazynina (approved):
this looks good to me.

routes/verify.go
maebeam: 200, empty response is the default

PR #85, opened 2021-07-22
Resend verification emails

maebeam:


routes/verify.go
```diff
@@ -364,10 +408,6 @@ func (fes *APIServer) VerifyEmail(ww http.ResponseWriter, req *http.Request) {
 }
 
 func (fes *APIServer) sendVerificationEmail(emailAddress string, publicKey string) {
-	if !fes.IsConfiguredForSendgrid() {
-		return
-	}
-
 	email := mail.NewV3Mail()
 	email.SetTemplateID(fes.Config.SendgridConfirmEmailId)
```
diamondhands0, line 412: What happens if no SendGrid key or whatever?

maebeam: IMO doesn't seem worthwhile to check for every flag every time we construct a specific time of email. I placed the checks at the endpoint level and in `sendEmail`

PR #95, opened 2021-07-31
Fix graceful shutdown of backend node

gfodor:
Noticed some segfaults when sending SIGTERM to backend. It appears that the `quit` channel is never created and the `APIServer` pointer on the node data structure was not being set.


tijno:
@gfodor im hopeful this will address this issue? 

https://github.com/bitclout/backend/issues/73

maebeam (approved):

maebeam:
Thank you!

PR #96, opened 2021-08-04
Ln/jumio

lazynina:


routes/global_state.go
```diff
@@ -212,6 +221,22 @@ type UserMetadata struct {
...
+
+	// JumioInternalReference = internal tracking reference for user's experience in Jumio
+	JumioInternalReference string
+	// JumioFinishedTime = has user completed flow in Jumio
+	JumioFinishedTime uint64
+	// JumioVerified = user was verified from Jumio flow
+	JumioVerified    bool
+	// JumioReturned = jumio webhook called
+	JumioReturned    bool
+	JumioTransactionID string
```
maebeam, line 233: Some additional comments on these values would be great

routes/server.go
```diff
@@ -1040,11 +1099,14 @@ func AddHeaders(inner http.Handler, allowedOrigins []string) http.Handler {
 		contentType := r.Header.Get("Content-Type")
 
 		invalidPostRequest := false
+		// TODO: add exception for application/x-www-form-urlencoded content-type for Jumio endpoint only.
```
maebeam, line 1102: Is this done?

routes/server.go
```diff
@@ -1040,11 +1099,14 @@ func AddHeaders(inner http.Handler, allowedOrigins []string) http.Handler {
...
 		// upload-image endpoint is the only one allowed to use multipart/form-data
 		if r.RequestURI == RoutePathUploadImage && strings.HasPrefix(contentType, "multipart/form-data") {
 			match = true
 			actualOrigin = "*"
-		} else if r.Method == "POST" && contentType != "application/json" {
+		//} else if r.Method == "POST" && contentType == "application/x-www-form-urlencoded" && r.RequestURI == RoutePathJumioCallback {
```
maebeam, line 1107: Remove?

routes/transaction.go
```diff
@@ -98,6 +98,37 @@ func (fes *APIServer) SubmitTransaction(ww http.ResponseWriter, req *http.Reques
 		return
 	}
 
+	// If this is a creator coin transaction and this node is configured for Jumio, we update global state user metadata to say user has purchased CC.
+	if txn.TxnMeta.GetTxnType() == lib.TxnTypeCreatorCoin && txn.TxnMeta.(*lib.CreatorCoinMetadataa).OperationType == lib.CreatorCoinOperationTypeBuy {
```
maebeam, line 102: I don't see the check for jumio or twilio?

routes/transaction.go
```diff
@@ -98,6 +98,37 @@ func (fes *APIServer) SubmitTransaction(ww http.ResponseWriter, req *http.Reques
...
+			userMetadata.HasPurchasedCreatorCoin = true
+			if err = fes.putUserMetadataInGlobalState(userMetadata); err != nil {
+				_AddBadRequestError(ww, fmt.Sprintf("SubmitTransactionRequest: Problem updating HasPurchasedCreatorCoin in global state user metadata: %v", err))
+				return
+			}
+		}
+	}
+
+	// If this is a basic transfer, we check if user has purchased CC (if this node is configured for Jumio or Twilio)
+	if txn.TxnMeta.GetTxnType() == lib.TxnTypeBasicTransfer && (fes.IsConfiguredForJumio() || fes.Twilio != nil){
```
maebeam, line 119: Why do we care if the node is configured for jumio or twilio? Should we set it regardless?

routes/verify.go
```diff
@@ -58,10 +61,21 @@ func (fes *APIServer) SendPhoneNumberVerificationText(ww http.ResponseWriter, re
 		return
 	}
 
+	// Validate their permissions
+	isValid, err := fes.ValidateJWT(requestData.PublicKeyBase58Check, requestData.JWT)
```
maebeam, line 65: Great

routes/verify.go
```diff
@@ -202,10 +222,21 @@ func (fes *APIServer) SubmitPhoneNumberVerificationCode(ww http.ResponseWriter,
...
```
maebeam, line 226: Excellent

routes/verify.go
```diff
@@ -452,3 +484,380 @@ func (fes *APIServer) sendEmail(email *mail.SGMailV3) {
...
+
+func (fes *APIServer) JumioBegin(ww http.ResponseWriter, req *http.Request) {
+	decoder := json.NewDecoder(io.LimitReader(req.Body, MaxRequestBodySizeBytes))
+	requestData := JumioBeginRequest{}
+	if err := decoder.Decode(&requestData); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("JumioBegin: Problem parsing request body: %v", err))
+		return
+	}
+
+	isValid, err := fes.ValidateJWT(requestData.PublicKey, requestData.JWT)
```
maebeam, line 527: Do we need check the err here? I saw you did above but those checks might be unnecessary. Can't imagine a case where isValid would be true and err would present

routes/verify.go
```diff
@@ -452,3 +484,380 @@ func (fes *APIServer) sendEmail(email *mail.SGMailV3) {
...
+	userMetadata.JumioReturned = false
+	if err = fes.putUserMetadataInGlobalState(userMetadata); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("JumioBegin: error putting jumio internal reference in global state: %v", err))
+		return
+	}
+
+	// CustomerInternalReference is Public Key + timestamp
+	// UserReference is just PublicKey
+	initData := &JumioInitRequest{
+		CustomerInternalReference: jumioInternalReference,
```
maebeam, line 565: Can we just use a UUID for this?

routes/verify.go
```diff
@@ -452,3 +484,380 @@ func (fes *APIServer) sendEmail(email *mail.SGMailV3) {
...
+
+	userMetadata.JumioFinishedTime = uint64(time.Now().UnixNano())
+
+	if err = fes.putUserMetadataInGlobalState(userMetadata); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("JumioFlowFinished: Error putting jumio finish time in global state: %v", err))
+		return
+	}
+}
+
+// Jumio webhook - If Jumio verified user is a human that we haven't paid already, pay them some starter CLOUT.
```
maebeam, line 654: Needs a comment about blocking non authorized IPs. We should probably figure out the best place to document this for both Wyre and Jumio. Deep in the code isn't a great spot.

routes/verify.go
```diff
@@ -452,3 +484,380 @@ func (fes *APIServer) sendEmail(email *mail.SGMailV3) {
...
+			_AddBadRequestError(ww, fmt.Sprintf("JumioCallback: Error putting user metdata in global state: %v", err))
+		}
+		return
+	}
+
+	// Make sure this id hasn't been verified before.
+	uniqueJumioKey := GlobalStateKeyForCountryIDDocumentTypeSubTypeDocumentNumber(idCountry, idType, idSubType, idNumber)
+	// We expect badger to return a key not found error if this document has not been verified before.
+	// If it does not return an error, this is a duplicate, so we skip ahead.
+	if val, _ := fes.GlobalStateGet(uniqueJumioKey); val == nil {
```
maebeam, line 751: Do we want to issue a Bad Request if it returns something? Will Jumio continue submitting callbacks if we error?

routes/verify.go
```diff
@@ -452,3 +484,380 @@ func (fes *APIServer) sendEmail(email *mail.SGMailV3) {
...
+		// to true and bail.
+		userMetadata.JumioReturned = true
+		if err = fes.putUserMetadataInGlobalState(userMetadata); err != nil {
+			_AddBadRequestError(ww, fmt.Sprintf("JumioCallback: Error putting user metdata in global state: %v", err))
+		}
+		return
+	}
+
+	// Make sure this id hasn't been verified before.
+	uniqueJumioKey := GlobalStateKeyForCountryIDDocumentTypeSubTypeDocumentNumber(idCountry, idType, idSubType, idNumber)
```
maebeam, line 748: I don't see us storing anything for this key?

routes/verify.go
```diff
@@ -452,3 +484,380 @@ func (fes *APIServer) sendEmail(email *mail.SGMailV3) {
...
+		// Update the user metadata to show that user has been jumio verified and store jumio transaction id.
+		userMetadata.JumioVerified = true
+		userMetadata.JumioReturned = true
+		userMetadata.JumioTransactionID = jumioTransactionId
+		userMetadata.JumioDocumentKey = uniqueJumioKey
+		userMetadata.JumioShouldCompProfileCreation = true
+		userMetadata.MustPurchaseCreatorCoin = true
+
+		bitcloutNanos := fes.GetJumioBitCloutNanos()
+		if bitcloutNanos > 0 {
```
maebeam, line 762: Don't we need a bad request error otherwise?

routes/verify.go
```diff
@@ -452,3 +484,380 @@ func (fes *APIServer) sendEmail(email *mail.SGMailV3) {
...
+			// Check the balance of the starter bitclout seed.
+			var balanceInsufficient bool
```
maebeam, line 764: Unnecessary var definition?

routes/verify.go
maebeam: Resubmit behavior is especially important if, for example, balance nanos is too low

config/config.go
```diff
@@ -61,6 +61,10 @@ type Config struct {
 	SendgridFromName       string
 	SendgridFromEmail      string
 	SendgridConfirmEmailId string
+
+	// Jumio
+	JumioToken             string
+	JumioSecret            string
```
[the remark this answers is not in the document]
lazynina: bitclout nanos is now managed in global state

config/config.go
```diff
@@ -135,5 +139,9 @@ func LoadConfig(coreConfig *coreCmd.Config) *Config {
 	config.SendgridFromEmail = viper.GetString("sendgrid-from-email")
 	config.SendgridConfirmEmailId = viper.GetString("sendgrid-confirm-email-id")
 
+	// Jumio
+	config.JumioToken = viper.GetString("jumio-token")
+	config.JumioSecret = viper.GetString("jumio-secret")
+
```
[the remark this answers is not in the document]
lazynina: nanos are now controlled by super admin in global state.

routes/verify.go
lazynina: should check error for completeness.

routes/verify.go
lazynina: we could but it significantly simplified things to have public key be a part of it when jumio flow finished stuff was happening in frontend instead of identity. I don't think it makes much of a difference either way now.

routes/verify.go
lazynina: Jumio will not submit again if we return an error

routes/verify.go
lazynina: if the amount of bitclout given for verifying with Jumio is 0, we just don't send bitclout. I don't think it requires an error necessarily.

routes/verify.go
maebeam: This sounds like a problem then in some error cases but not in others?

routes/verify.go
maebeam: Gotcha, misread

routes/transaction.go
lazynina: yep - I'll update the comment.

routes/transaction.go
lazynina: it would make a node run faster if they didn't fetch the user metadata for every basic transfer

lazynina: Basically, the usermetadata would never have a phone number that isn't the empty string nor would jumio verified ever be true, so it saves a trip to global state.

routes/admin_jumio.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	utxoView, err := fes.backendServer.GetMempool().GetAugmentedUniversalView()
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("AdminResetJumioForPublicKey: error getting utxoview: %v", err))
+		return
+	}
+
+	var userMetadata *UserMetadata
+	if requestData.PublicKeyBase58Check != "" {
```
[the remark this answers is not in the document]
lazynina: no? We are fetching userMetadata here which we don't do in other places where we check public key or username

routes/verify.go
lazynina: user can always go through Jumio flow again since we won't mark them as jumio verified. worst case scenario is that we give them money and then something fails when we are putting the user's data in global state - this error should occur very rarely. If we get an error because the balance is too low to cover the starter clout amount, the user can always go through the flow later. We should just monitor the balance that send CLOUT here and make sure it's always got enough juice. Setting up a retry queue for this will be a bit annoying.

routes/verify.go
lazynina: Yeah I definitely killed this during testing, but I've added it back now. thanks

cmd/run.go
```diff
@@ -137,6 +137,10 @@ func init() {
 	runCmd.PersistentFlags().String("sendgrid-from-email", "", "Sendgrid from email")
 	runCmd.PersistentFlags().String("sendgrid-confirm-email-id", "", "Sendgrid confirmation email template ID")
 
+	runCmd.PersistentFlags().String("jumio-token", "", "Jumio Token")
+	runCmd.PersistentFlags().String("jumio-secret", "", "Jumio Secret Key")
+	runCmd.PersistentFlags().Uint64("jumio-bitclout-nanos", 0, "BitClout sent for verifying with Jumio")
```
maebeam, line 142: Remove this flag if it's unused

routes/admin_jumio.go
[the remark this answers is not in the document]
maebeam: Ah I thought there was another place where we get metadata for username or public key

maebeam (approved):

routes/verify.go
```diff
@@ -452,3 +484,391 @@ func (fes *APIServer) sendEmail(email *mail.SGMailV3) {
...
+
+func (fes *APIServer) JumioFlowFinished(ww http.ResponseWriter, req *http.Request) {
+	decoder := json.NewDecoder(io.LimitReader(req.Body, MaxRequestBodySizeBytes))
+	requestData := JumioFlowFinishedRequest{}
+	if err := decoder.Decode(&requestData); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("JumioFlowFinished: Problem parsing request body: %v", err))
+		return
+	}
+
+	// Get UserMetadata from global state and check internal reference matches and we haven't marked this user as finished already.
```
maebeam, line 638: Missing JWT validation

PR #98, opened 2021-08-06
feat: Added new route to enable showcase preview panel in admin tab. …

superzordon:
…Also added uniqueness constraint to showcase drop post hashes


routes/admin_nft.go
```diff
@@ -295,8 +305,13 @@ func (fes *APIServer) AdminUpdateNFTDrop(ww http.ResponseWriter, req *http.Reque
...
 			}
-
-			updatedDropEntry.NFTHashes = append(updatedDropEntry.NFTHashes, postHash)
+			if BlockHashIn(updatedDropEntry.NFTHashes, postHash) {
+				_AddBadRequestError(ww, fmt.Sprintf(
+					"AdminUpdateNFTDrop: Cannot add duplicate NFT to drop: %v", postHash.String()))
+				return
+			} else {
+				updatedDropEntry.NFTHashes = append(updatedDropEntry.NFTHashes, postHash)
+			}
```
lazynina, line 314:
```suggestion
			}
			updatedDropEntry.NFTHashes = append(updatedDropEntry.NFTHashes, postHash)
```

Since we return above, we don't need an else.

routes/nft.go
```diff
@@ -694,14 +693,53 @@ func (fes *APIServer) GetNFTShowcase(ww http.ResponseWriter, req *http.Request)
...
+	requestData := GetNFTShowcaseRequest{}
+	if err := decoder.Decode(&requestData); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("GetNFTShowcase: Error parsing request body: %v", err))
+		return
+	}
+	fes.GetNFTShowcase(ww, req, -1, requestData.ReaderPublicKeyBase58Check)
+}
+
+// If dropIdx == -1, get current drop
+func (fes *APIServer) GetNFTShowcase(ww http.ResponseWriter, req *http.Request, dropIdx int, readerPublicKeyBase58Check string) {
```
lazynina, line 717: do we need `req *http.Request` as a param in this function?  doesn't look like it's used.

routes/shared.go
```diff
@@ -339,3 +339,12 @@ func (fes *APIServer) SendSeedBitClout(recipientPkBytes []byte, amountNanos uint
 	}
 	return hash, err
 }
+
+func BlockHashIn(slice []*lib.BlockHash, val *lib.BlockHash) bool {
+	for _, item := range slice {
+		if item.String() == val.String() {
```
lazynina, line 345: for comparing block hashes, we typically use reflect.DeepEqual. It may be a bit overkill, but we should be consistent.

lazynina:
```suggestion
		if reflect.DeepEqual(item, val) {
```

scripts/nodes/n0_test
```diff
@@ -30,7 +30,7 @@ rm /tmp/main.*.log
   --protocol-port=18000 \
   --txindex \
   --num-mining-threads=1 \
-  --miner-public-keys=BC1YLg7Bk5sq9iNY17bAwoAYiChLYpmWEi6nY6q5gnA1UQV6xixHjfV \
+  --miner-public-keys=tBCKW665XZnvVZcCfcEmyeecSZGKAdaxwV2SH9UFab6PpSRikg4EJ2 \
```
lazynina, line 33: danke. maybe add a comment that says that's the public key associated with the starter-bitclout-seed below?

lazynina (changes requested):

routes/nft.go
```diff
@@ -656,36 +645,46 @@ func (fes *APIServer) GetNFTShowcase(ww http.ResponseWriter, req *http.Request)
...
 	if err != nil {
-		_AddBadRequestError(ww, fmt.Sprintf("GetNFTBidsForNFTPost: Error getting verified user map: %v", err))
-		return
+		return nil, fmt.Errorf(
+			"GetNFTBidsForNFTPost: Error getting verified user map: %v", err)
 	}
 
 	var readerPKID *lib.PKID
-	if requestData.ReaderPublicKeyBase58Check != "" {
+	if ReaderPublicKeyBase58Check != "" {
```
lazynina, line 679: I don't think we need to pass the ReaderPublicKeyBase58Check in here. we can instead check that the length of the readerPublicKeyBytes is correct.

`if len(readerPublicKeyBytes) == btcec.PubKeyBytesLenCompressed`

let's remove ReaderPublicKeyBase58Check from this funciton

lazynina (changes requested):
Think if we remove ReaderPublicKeyBase58Check from `FetchNFTSForDrop` then this will be good to go

PR #105, opened 2021-08-12
remove some jumio global state items

lazynina:


maebeam (approved):

routes/verify.go
```diff
@@ -825,7 +807,7 @@ func (fes *APIServer) JumioCallback(ww http.ResponseWriter, req *http.Request) {
 			// Save transaction hash hex in user metadata.
 			userMetadata.JumioStarterBitCloutTxnHashHex = txnHash.String()
 		}
-		if err = fes.GlobalStatePut(uniqueJumioKey, payloadBytes); err != nil {
+		if err = fes.GlobalStatePut(uniqueJumioKey, []byte{}); err != nil {
```
netpro2k, line 810: I had brought this up as an issue so good to see that the jumio response is no longer being stored as it contained some sensitive info. 

The document id still appears to be part of the `uniqueJumioKey` which means sensitive info is still being stored. Maybe it could be hashed? This would allow preventing duplicates without needing to store the actual id in the database. Lookup would be slightly annoying since you would need to check both hashed and unhahsed for legacy support, or do a one time migration. (migration might not be a bad idea so that all old entries that were storing the whole response can also be cleared.

seanslater: What is the sensitive information stored?

Is it something that users that have verified before this change need to be aware of?

maebeam: There's no real advantage to hashing the document ID because the search space is so small. All the hashes could be back computed instantly.

maebeam: The document ID is not associated with a specific user so it's pretty much useless. We will be running a migration to remove old entries.

netpro2k: Fair point on the keyspace at least for the ID numbers I am familiar with (US passport and state id numbers tend to be ~9 digit numeric ids). This space could be expanded a bit if you include some of the other properties in the response, but is still likely pretty small. It could be made at least user specific by including the bitclout public key as part of the hash....

That said, I had missed that this PR also removed setting the `JumioDocumentKey` on the userMetadata object, so like you said it's not associated with any particular account anymore so I guess the point is pretty moot.

I did notice that the `JumioTransactionID` is still being kept. With that it looks like its possible to recover the same response information this PR removes (https://github.com/Jumio/implementation-guides/blob/master/netverify/netverify-retrieval-api.md#retrieving-details). There is a level of indirection since you would also need the Jumio API secret but it still seems undesirable to be able to recover that. Is this by design? If so that should probably be communicated somehow in the verification flow.

PR #111, opened 2021-08-17
Onboarding Tutorial

lazynina:


maebeam (approved):
My only concern is users who currently have `MustPurchaseCreatorCoin = true` will not be forced to complete the tutorial

routes/user.go
```diff
@@ -1078,6 +1104,18 @@ func (fes *APIServer) GetSingleProfile(ww http.ResponseWriter, req *http.Request
 		res.IsGraylisted = true
 	}
 
+	if isAdmin, _ := fes.UserAdminStatus(publicKeyBase58Check); isAdmin {
```
maebeam, line 1107: Why are we checking if the profile we're fetching is an admin? This seems wrong?

lazynina: Basically, we only need the IsFeaturedTutorialUpAndComing and IsFeaturedTutorialWellKnownCreator values if the user is an admin. These values are only used to determine whether to show "Add" or "Remove" from the well-known and up-and-coming creator indexes on a creator's profile. We don't need to fetch user metadata from global state for anything else, so if we're not an admin viewing a profile, it's a little bit faster.

lazynina:
> My only concern is users who currently have `MustPurchaseCreatorCoin = true` will not be forced to complete the tutorial

I thought that as well, but it felt simpler to remove that MustPurchaseCreatorCoin stuff all together. Happy to add it back in if we decide we want to enforce it. Will defer to @diamondhands0 on that though

routes/user.go
lazynina: oh I see - yes we need to update this to take in the reader public key. Good catch!

redpartyhat (approved):

diamondhands0 (approved):

PR #133, opened 2021-08-31
Performance enhancements

lazynina:
- move getVerifiedUsernameToPKIDMap calls outside of loops and helper functions
- remove unnecessary call to get the profile entry response in augmentProfileEntry


superzordon (approved):
🔥

PR #136, opened 2021-09-01
Rph/referrals

lazynina:


maebeam (approved):

routes/verify.go
```diff
@@ -800,50 +820,134 @@ func (fes *APIServer) JumioCallback(ww http.ResponseWriter, req *http.Request) {
...
+	userMetadata.RedoJumio = false
+
+	if bitcloutNanos := fes.GetJumioBitCloutNanos(); bitcloutNanos > 0 || userMetadata.ReferralHashBase58Check != "" {
+		payReferrer := false
+		if userMetadata.ReferralHashBase58Check != "" {
+			// TODO: pay the referrer
+			var referralInfo *ReferralInfo
+			referralInfo, err = fes.getInfoForReferralHashBase58(userMetadata.ReferralHashBase58Check)
+			if err != nil {
+				glog.Errorf("JumioVerifiedHandler: Error getting referral info: %v", err)
```
diamondhands0, line 858: nit: There is no need to do "if else" when checking. Instead it's better to do:
```
if err != nil {
  // whatever
}

// rest of code
```

routes/verify.go
```diff
@@ -800,50 +820,134 @@ func (fes *APIServer) JumioCallback(ww http.ResponseWriter, req *http.Request) {
...
+			if err = fes.putReferralHashWithInfo(userMetadata.ReferralHashBase58Check, referralInfo); err != nil {
+				return userMetadata, fmt.Errorf("JumioVerifiedHandler: Error updating referral info. Skipping paying referrer: %v", err)
+			}
+			// Bail if there referrer gets nothing
+			if referrerBitcloutNanos == 0 {
+				return userMetadata, nil
+			}
+			// Send the referrer money
+			var referrerTxnHash *lib.BlockHash
+			referrerTxnHash, err = fes.SendSeedBitClout(referrerPublicKeyBytes, referrerBitcloutNanos, false)
```
diamondhands0, line 942: How do we check the case where the same person passes Jumio twice? I.e. how do we prevent the referrer from getting paid twice in that case?

lazynina: Just like it behaves now - we only pay the user if we haven't encountered their ID in jumio before or a super admin has reset the jumio state for the user. If the ID is not new, we exit early.

PR #141, opened 2021-09-15
Remove hardcoded base58 key prefixes and compute based on params

gfodor:
Several places in `backend` hardcoded the public key base 58 prefix `BC` and `tBC`. This PR updates the code to compute this prefix dynamically based on `Params`, which can be useful if the prefix has been adjusted (for example, in a private testnet.)


maebeam (approved):

maebeam:
While this change looks like good cleanup, can I ask why you're changing the prefix? I believe there are many other places that assume a specific prefix or key length

gfodor:
I’m experimenting with running a private network with a different key space. Core seemed to work fine without any changes other than configs but hit this snag trying to create transactions via backend.

PR #142, opened 2021-09-15
Buy clout with eth

maebeam:


routes/eth.go
```diff
@@ -0,0 +1,552 @@
...
+
+import (
+	"bytes"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"github.com/bitclout/core/lib"
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcutil"
+	"github.com/golang/glog"
```
redpartyhat, line 11: I had to kill this line in order to get ./n0_test to run.

routes/eth.go
```diff
@@ -0,0 +1,552 @@
...
+		},
+		Outputs: []BlockCypherTxOutput{
+			{
+				Addresses: []string{
+					fes.Config.BuyBitCloutETHAddress[2:], // Remove the 0x prefix
+				},
+				Value: amount,
+			},
+		},
+	})
```
redpartyhat, line 337: Is there any way to prevent replace by fee? It would be bad if someone could buy clout with ETH and then broadcast a transaction to cancel their ETH txn: https://fixedfloat.com/en/blog/guides/replace-by-fee-ethereum#:~:text=What%20is%20the%20Replace%20by,after%20launch%20are%20considered%20unconfirmed.

maebeam: Hmmm, that's pretty bad. We probably need to wait for a block then

redpartyhat (approved):

PR #148, opened 2021-09-22
Ln/native video

lazynina:
- Add flags for cloudflare accountId and token
- Add endpoints for uploading a video using a [resumable upload](https://developers.cloudflare.com/stream/uploading-videos/direct-creator-uploads#using-tus-recommended-for-videos-over-200mb)
- Add endpoint for checking if a video is ready to be streamed
- Update PostEntryResponse struct to have VideoURLs as an attribute
- Update the cleanBody method used to validate the body for a post to include VideoURLs


maebeam (approved):
Looks great!

routes/media.go
```diff
@@ -296,3 +297,93 @@ func (fes *APIServer) GetFullTikTokURL(ww http.ResponseWriter, req *http.Request
 		return
 	}
 }
+
+func (fes *APIServer) UploadVideo(ww http.ResponseWriter, req *http.Request) {
```
redpartyhat, line 301: It's not clear to me what this endpoint does exactly.  Would you mind adding a comment above it to make it more clear?

routes/media.go
```diff
@@ -296,3 +297,93 @@ func (fes *APIServer) GetFullTikTokURL(ww http.ResponseWriter, req *http.Request
...
+		_AddBadRequestError(ww, fmt.Sprintf("UploadVideo: This node is not configured to support video uploads"))
+		return
+	}
+	url := fmt.Sprintf("https://api.cloudflare.com/client/v4/accounts/%v/stream?direct_user=true", fes.Config.CloudflareAccountId)
+	client := &http.Client{}
+
+	request, err := http.NewRequest("POST", url, nil)
+	request.Header.Add("Authorization", fmt.Sprintf("Bearer %v", fes.Config.CloudflareStreamToken))
+	request.Header.Add("Tus-Resumable", "1.0.0")
+	request.Header.Add("Upload-Length", req.Header.Get("Upload-Length"))
```
redpartyhat, line 312: Is there any check on video upload size? Do we need one?

lazynina: Yea we can check the Upload-Length here as well in the frontend. What do you think would be a reasonable size for video uploads? Youtube supports up to 256GB (which seems excessive) I was thinking something in the 16GB to 64GB range as a limit.

routes/media.go
lazynina: Added some comments - let me know if it's more clear now.

routes/media.go
lazynina: Added a 4GB limit here and on the frontend. We can make it configurable later.

routes/media.go
redpartyhat: Loooooooooooove it! Thanks :)

redpartyhat (approved):

PR #150, opened 2021-09-24
Add admin eth processing tool

maebeam:


redpartyhat (approved):

routes/eth.go
```diff
@@ -241,41 +260,124 @@ func (fes *APIServer) SubmitETHTx(ww http.ResponseWriter, req *http.Request) {
...
+	if err := decoder.Decode(&requestData); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("AdminProcessETHTx: Problem parsing request body: %v", err))
 		return
 	}
 
-	// Record transaction success in global state
-	if err = fes.GlobalStatePut(globalStateKey, []byte{1}); err != nil {
-		_AddBadRequestError(ww, fmt.Sprintf("SubmitETHTx: Error processing GlobalStatePut: %v", err))
+	if !fes.IsConfiguredForETH() {
+		_AddBadRequestError(ww, "AdminProcessETHTx: Not configured for ETH")
```
lazynina, line 338: I would move this ahead of the decoding since we won't even want to bother decoding the payload if the node is not configured for ETH

routes/eth.go
```diff
@@ -241,41 +260,124 @@ func (fes *APIServer) SubmitETHTx(ww http.ResponseWriter, req *http.Request) {
...
+	bitcloutTxnHash, err := fes.SendSeedBitClout(ethTxLog.PublicKey, nanosPurchased, true)
+	if err != nil {
+		return nil, errors.New(fmt.Sprintf("Error sending BitClout: %v", err))
+	}
+
+	// Record successful transaction in global state
+	ethTxLog.BitCloutTxnHash = bitcloutTxnHash.String()
+	globalStateKey := GlobalStateKeyETHPurchases(ethTxLog.Tx.Hash)
+	globalStateVal := bytes.NewBuffer([]byte{})
+	err = gob.NewEncoder(globalStateVal).Encode(ethTxLog)
```
lazynina, line 308: In the Wyre flow, we mark the transaction as paid out before we send money and then unmark it if there is an error sending the money. If I recall correctly, the rationale was that we won't want someone to be able to continually get money if there's an issue with global state for some reason. I think it's okay here, but just wanted to mention it here.

routes/eth.go
```diff
@@ -241,41 +260,124 @@ func (fes *APIServer) SubmitETHTx(ww http.ResponseWriter, req *http.Request) {
...
+	}
+
+	// Transaction must be unsuccessful
+	if len(ethTxLog.BitCloutTxnHash) > 0 {
+		_AddBadRequestError(ww, fmt.Sprintf("AdminProcessETHTx: CLOUT was sent: %s", ethTxLog.BitCloutTxnHash))
+		return
+	}
+
+	// Fetch the transaction
+	ethTx, err := fes.BlockCypherGetETHTx(requestData.ETHTxHash)
```
lazynina, line 364: Is there ever a possibility that ethTx will be non-nil and error will be nil? Should we update the condition to be `if ethTx == nil || err != nil {`

lazynina (changes requested):

routes/eth.go
maebeam: Fixed! Ty

PR #159, opened 2021-09-29
Add back deprecated bitclout fields for backwards compatability

maebeam:


AeonSw4n:
LGTM )

chafreaky:
Will this include all previous responses or only the ones addressed so far?

maebeam:
@chafreaky what responses am I missing? I tried to get all of them

chafreaky:
@maebeam specifically the CoinEntry ones (I see you took care of some) - I don't see one for "DeSoLockedNanos". Might be missing it though.

maebeam:
Thanks! Just added support for `BitCloutLockedNanos`

routes/post.go
```diff
@@ -207,8 +208,8 @@ func (fes *APIServer) _postEntryToResponse(postEntry *lib.PostEntry, addGlobalFe
 		LikeCount:                      postEntry.LikeCount,
 		DiamondCount:                   postEntry.DiamondCount,
 		CommentCount:                   postEntry.CommentCount,
-		RepostCount:                   postEntry.RepostCount,
-		QuoteRepostCount:              postEntry.QuoteRepostCount,
+		RepostCount:                    postEntry.RepostCount,
```
lazynina, line 211: We should add back RecloutCount and QuoteRecloutCount to maintain backwards compatibility

lazynina (changes requested):
Can we also add back routes that have replaced bitclout or clout with deso in the path? There are just a few I think - mainly in admin_buy_deso

routes/post.go
maebeam: Ty!

PR #160, opened 2021-09-30
Ln/node monetization

lazynina:


routes/admin_fees.go
```diff
@@ -0,0 +1,210 @@
...
+	"encoding/json"
+	"fmt"
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/deso-protocol/core/lib"
+	"github.com/golang/glog"
+	"io"
+	"net/http"
+)
+
+type TransactionFee struct {
```
redpartyhat, line 15: Let's add some comments to the struct fields here. In particular:
1) PublicKeyBase58Check --> Specify that this is the fee recipient.  At first I thought these fees were only applicable to specific public keys.
2) ProfileEntryResponse --> It looks like the ProfileEntryResponse is just used when the TransactionFee is part of an endpoint's Response but is typically nil when the frontend provides TransactionFees as part of a Request.  Would be helpful to clarify so that someone who is interacting with this API doesn't feel like they need to go dig up a ProfileEntryResponse in order to have it work properly.

redpartyhat (approved):

maebeam:
I think this needs more safeguards. It becomes too easy for anyone who gains admin access to instantly steal everyone's money. Not sure how to implement this best

routes/admin_fees.go
```diff
@@ -374,3 +385,18 @@ func (fes *APIServer) GetExemptPublicKeyMapFromGlobalState() map[string]interfac
...
+func (fes *APIServer) LogFeeSet(txnType lib.TxnType, transactionFees []TransactionFee) (_err error) {
+	if fes.backendServer == nil || fes.backendServer.GetStatsdClient() == nil {
+		return nil
+	}
+	var totalFees uint64
+	for _, transactionFee := range transactionFees {
+		totalFees += transactionFee.AmountNanos
+	}
+	tags := []string{}
+	if err := fes.backendServer.GetStatsdClient().Gauge(fmt.Sprintf("%v_NODE_FEE_TOTAL", txnType.GetTxnString()), float64(totalFees), tags, 1); err != nil {
```
maebeam, line 398: Can we please make the format `NODE_FEE.%v`

PR #161, opened 2021-10-01
Change admin public key default behavior

redpartyhat:
At the moment, if a node spins up without admin public keys, any user is able to access that nodes admin API.  This has led to a security incident in the past and doesn't seem like the best default behavior.

In this PR, I replace this behavior with a new flag, "force-ignore-admin-public-keys".  This flag has to be explicitly set to true for nodes to behave in a way that allows public access to the admin API.


lazynina (approved):
LGTM

diamondhands0:
Why a new flag rather than allowing someone to pass a string like “ignore” to the old flag?

lazynina:
> Why a new flag rather than allowing someone to pass a string like “ignore” to the old flag?

The old logic goes as follows:
If neither --admin-public-keys nor --super-admin-public-keys are set, all users are admins and super admins.

This is dangerous as a simple mistake with a configuration would allow any user to perform protected actions - most notably with the addition of the node monetization fee, as the user would be able to add a fee to all transactions on the node.

The change proposed here would alter the default behavior. If neither admin key flag is specified, then by default, nobody is an admin. If you wish to have the old behavior, then you can set the flag to true.

maebeam:
I prefer @diamondhands0's suggestion of modifying the existing flag to take something like "ALL" and make everyone an admin. An empty list means nobody is an admin.

redpartyhat:
Updated! The new behavior is: you must explicitly specify "admin-public-keys=*" or "super-admin-public-keys=*" in order to give anyone access to either set of functionality.

maebeam (approved):
Fantastic change!

tijno (changes requested):
As the original purpose/idea of this change was to prevent the bad default of having admin open for new installs - should default for non test scripts be to leave admin/super admin empty / closed, so remove the `*` ?

And similar in run repo - maybe comment out the admin/super keys so user specifically has to uncomment line to enable admins access for all or some?

maebeam:
The idea is that if you mess up the script / environment variable that your node doesn't suddenly reveal its admin panel to the world. Having the default set to `*` is still the desired behavior for users spinning up a node for the first time.

maebeam:
Merge away!

PR #183, opened 2021-10-19
Fix GetNFTShowcase endpoint iterating over burned NFTs

kanshi:
Showcase was hard-failing when it included an NFT with all copies burnt.


maebeam:
Switch it to a continue and use tabs instead of spaces. Then I can merge :)

PR #192, opened 2021-10-26
Merge this node's global state in with remote global state for verifications, blacklist, graylist and Global feed

lazynina:
I've kept this separate from the existing work to expose global state as it changes how we remove verification from a user.

Additionally, this change would limit the global feed to only the latest week of posts. I think this is a far constraint to apply in order to improve performance. We could extend that window further back if we think that's too tight.


routes/server.go
```diff
@@ -1916,10 +1920,11 @@ func (fes *APIServer) SetGlobalStateCache() {
 		glog.Errorf("SetGlobalStateCache: problem with GetAugmentedUniversalView: %v", err)
 		return
 	}
+
```
lazynina, line 1923:
```suggestion
```

PR #201, opened 2021-11-03
Z/notifications alert

superzordon:
Add new endpoint to count a user's unread notifications. Also split out current notifications function into sub-functions (getDBNotifications & getMempoolNotifications).

### DIFF
--- a/routes/global_state.go
+++ b/routes/global_state.go
@@ -345,6 +345,11 @@ type UserMetadata struct {
 
 	// Txn hash in which the referrer was paid
 	ReferrerDeSoTxnHash string
+
+	// The number of unread notifications stored in the db.
+	UnreadNotifications uint64
+	// The most recently scanned notification transaction index in the database. Stored in order to prevent unnecessary re-scanning.
+	LatestUnreadNotificationIndex int64
 }
 
 type TutorialStatus string
--- a/routes/user.go
+++ b/routes/user.go
@@ -2195,7 +2195,7 @@ func (fes *APIServer) _getDBNotifications(request *GetNotificationsRequest, bloc
 	return dbTxnMetadataFound, nil
 }
 
-func (fes *APIServer) _getMempoolNotifications(request *GetNotificationsRequest, blockedPubKeys map[string]struct{}, utxoView *lib.UtxoView) ([]*TransactionMetadataResponse, error) {
+func (fes *APIServer) _getMempoolNotifications(request *GetNotificationsRequest, blockedPubKeys map[string]struct{}, utxoView *lib.UtxoView, iterateReverse bool) ([]*TransactionMetadataResponse, error) {
 	filteredOutCategories := request.FilteredOutNotificationCategories
 
 	pkBytes, _, err := lib.Base58CheckDecode(request.PublicKeyBase58Check)
@@ -2214,7 +2214,7 @@ func (fes *APIServer) _getMempoolNotifications(request *GetNotificationsRequest,
 	// than what we have in the db then it means we need to augment our list with
 	// txns from the mempool.
 	combinedMempoolDBTxnMetadata := []*TransactionMetadataResponse{}
-	if request.FetchStartIndex < 0 || request.FetchStartIndex >= NextIndex {
+	if !iterateReverse || (request.FetchStartIndex < 0 || request.FetchStartIndex >= NextIndex) {
 		// At this point we should have zero or more TransactionMetadata objects from
 		// the database that could trigger a notification for the user.
 		//
@@ -2267,7 +2267,7 @@ func (fes *APIServer) _getMempoolNotifications(request *GetNotificationsRequest,
 				}
 
 				// Only include transactions that occur on or after the start index, if defined
-				if request.FetchStartIndex < 0 || request.FetchStartIndex >= currentIndex {
+				if request.FetchStartIndex < 0 || (request.FetchStartIndex >= currentIndex && iterateReverse) || (request.FetchStartIndex <= currentIndex && !iterateReverse)  {
 					mempoolTxnMetadata = append(mempoolTxnMetadata, &TransactionMetadataResponse{
 						Metadata: txnMeta,
 						Index:    currentIndex,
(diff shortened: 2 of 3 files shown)


routes/global_state.go
```diff
@@ -345,6 +345,11 @@ type UserMetadata struct {
 
 	// Txn hash in which the referrer was paid
 	ReferrerDeSoTxnHash string
+
+	// The number of unread notifications stored in the db.
+	UnreadNotifications uint64
+	// The most recently scanned notification transaction index in the database. Stored in order to prevent unnecessary re-scanning.
+	LatestUnreadNotificationIndex int64
```
redpartyhat, line 352: There is an index in this struct called, "NotificationLastSeenIndex".  What's the difference between these two variables?  If it's necessary to have both, we should probably clarify or mark one as deprecated (unless it is used in the bitclout.com frontend).

routes/user.go
```diff
@@ -1811,12 +1811,87 @@ type GetNotificationsRequest struct {
...
+
+func (fes *APIServer) GetNotificationsCount(ww http.ResponseWriter, req *http.Request) {
+	decoder := json.NewDecoder(io.LimitReader(req.Body, MaxRequestBodySizeBytes))
+	requestData := GetNotificationsCountRequest{}
+	if err := decoder.Decode(&requestData); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf(
+			"GetNotificationsCount: Problem parsing request body: %v", err))
+		return
+	}
+
```
redpartyhat, line 1837: We should probably validate the JWT here to make sure that the person requesting the notifications count for the given public key is in fact the owner of that public key.  Notification count is not super sensitive PII but people might still want to keep it private.  See the referral info endpoints for an example of how to validate the JWT is signed by the correct pub key: https://github.com/deso-protocol/backend/blob/ff3b82bdeca8a469dfeb03cb09254b6daa3cbfd5/routes/referrals.go#L32-L41

routes/user.go
```diff
@@ -1811,12 +1811,87 @@ type GetNotificationsRequest struct {
...
+	var fetchStartIndex int64
+
+	if userMetadata.LatestUnreadNotificationIndex > 0 {
+		fetchStartIndex = userMetadata.LatestUnreadNotificationIndex + 1
+	} else if userMetadata.NotificationLastSeenIndex > 0 {
+		fetchStartIndex = userMetadata.NotificationLastSeenIndex + 1
+	}
```
redpartyhat, line 1851: Oh, it's interesting that you are using both of these.  Why do we need both?

routes/user.go
```diff
@@ -1811,12 +1811,87 @@ type GetNotificationsRequest struct {
...
+
+	// Update user metadata with new unread notification count.
+	userMetadata.UnreadNotifications = notificationsCount
+	// Save the new latest unread notification index, so that notifications aren't scanned twice. Only do this when new notifications are added.
+	if newNotificationsCount > 0 {
+		userMetadata.LatestUnreadNotificationIndex = newNotificationsStartIndex
+	}
+
+	if err = fes.putUserMetadataInGlobalState(userMetadata); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("AdminResetTutorialStatus: Error putting user metadata in global state: %v", err))
```
redpartyhat, line 1879: replace AdminResetTutorialStatus w/GetNotificationsCount

routes/user.go
```diff
@@ -1811,12 +1811,87 @@ type GetNotificationsRequest struct {
...
+		return
+	}
+
+	res := &GetNotificationsCountResponse{
+		NotificationsCount: notificationsCount,
+	}
+
+	if err := json.NewEncoder(ww).Encode(res); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf(
+			"GetNotifications: Problem encoding response as JSON: %v", err))
```
redpartyhat, line 1889: nit: replace GetNotifications w/GetNotificationsCount

routes/user.go
```diff
@@ -1811,12 +1811,87 @@ type GetNotificationsRequest struct {
 	FilteredOutNotificationCategories map[string]bool
 }
 
+type GetNotificationsCountRequest struct {
```
redpartyhat, line 1814: nit: It would be great if we could move the "GetNotificationsRequest" and "GetNotificationsResponse" structs back down so that they immediately precede the GetNotfications() handler.  Most endpoint handler functions in the backend repo follow this pattern and it's super convenient when you are reading the code.

routes/user.go
```diff
@@ -1811,12 +1811,87 @@ type GetNotificationsRequest struct {
...
+		_AddBadRequestError(ww, fmt.Sprintf(
+			"GetNotifications: Problem encoding response as JSON: %v", err))
+		return
+	}
+}
+
+
 func (fes *APIServer) GetNotifications(ww http.ResponseWriter, req *http.Request) {
 	decoder := json.NewDecoder(io.LimitReader(req.Body, MaxRequestBodySizeBytes))
 	requestData := GetNotificationsRequest{}
```
redpartyhat, line 1897: This function needs to validate that the JWT is signed by the public key that notifications are being requested for (see comment on the GetNotificationsCount function).

routes/global_state.go
superzordon: `NotificationLastSeenIndex` tracks the last notification that a user has read, and allows the backend to determine which notifications are therefore "unread"

`LatestUnreadNotificationIndex` tracks the last index that the backend has scanned via the `getUnreadNotifications` route. By tracking this index, we can ensure that we don't duplicate work, and that we only scan new notifications once for this route.

routes/user.go
superzordon: If we've already run `getUnreadNotificationsCount` before, we want to start where we left off last time so as to not duplicate work (the result from the last scan will be stored in `UnreadNotifications`). If `getUnreadNotificationsCount` has not been called since the user last checked their notifications, we want to continue from the last notification the user has read.

routes/user.go
superzordon: Logic that effects a user's metadata has been moved to a separate route that is JWT protected.

routes/user.go
superzordon: Structs have been moved

redpartyhat (approved):

diamondhands0 (approved):

PR #208, opened 2021-11-09
Remove unused config values

maebeam:


lazynina (approved):
:+1: love to see it.

lazynina (commented):
~We need the datadog profiler in core currently.~

lazynina (changes requested):
I think we can leave the datadog profiler for now since it is used in core - otherwise this looks good.

lazynina (approved):
Core has its own config - this is good to go.

PR #212, opened 2021-11-17
Fix wildcard behavior for admin public keys

maebeam:
When running a global state node we do not send the AdminPublicKey param and we were not making it to the "*" check. This PR moves that check to the very start of the method.


lazynina (approved):
LGTM

PR #221, opened 2021-11-29
add default tx fees to app state

tijno:
This PR adds the default Fees to Get App State - which seems sensible as it returns already Create Profile fees as well as Node Tx Fees.


tijno:
Set this to draft as I am not in a position to test this right now. If anyone else can - feel free. 

cc @iPaulPro @brootle

maebeam (approved):

PR #222, opened 2021-11-29
add submit-transaction endpoint to list of public routes

lazynina:
In order to support profile creation compensation + submitting the transaction to this node, we need to expose this route.


AeonSw4n (approved):
Yess!

maebeam (approved):

PR #226, opened 2021-12-04
validate the transaction passed in instead of fetching from API again

lazynina:
Why make this change?

There was an instance where the ETH tx mined, so we broke out of the loop that polls Infura's API, but then in FinishETHTx, the call to get the tx from Infura's API returned a tx without a block number, so we aborted.

This change will prevent this discrepancy.

### DIFF
--- a/routes/eth.go
+++ b/routes/eth.go
@@ -154,7 +154,7 @@ func (fes *APIServer) SubmitETHTx(ww http.ResponseWriter, req *http.Request) {
 			break
 		}
 	}
-	// The transaction has mined so we finish by validating the transaction again and paying the user.
+	// The transaction has mined or we've waited for 10 minutes so we finish by validating the transaction and paying the user.
 	desoTxHash, err := fes.finishETHTx(ethTx, ethTxLog)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


routes/eth.go
```diff
@@ -171,24 +171,17 @@ func (fes *APIServer) SubmitETHTx(ww http.ResponseWriter, req *http.Request) {
...
-	ethTx, err := fes.GetETHTransactionByHash(ethTxIn.Hash)
-	if err != nil  {
-		return nil, errors.New(fmt.Sprintf("Failed to get eth transaction: %v", err))
-	}
-
-	// Ensure the transaction mined
-	if ethTx.BlockNumber == nil {
+func (fes *APIServer) finishETHTx(ethTx *InfuraTx, ethTxLog *ETHTxLog) (desoTxHash *lib.BlockHash, _err error) {
+	if ethTx == nil || ethTx.BlockNumber == nil {
 		return nil, errors.New("Transaction failed to mine")
```
diamondhands0, line 180: I'd log the eth txn if we have it so we can see it in logs.

diamondhands0 (approved):

maebeam (approved):

PR #239, opened 2021-12-10
Seperate out global state into it's own struct in order for it to be …

superzordon:
…seperately importable from other repositories


routes/expose_global_state.go
```diff
@@ -1,4 +1,4 @@
-package routes
+	package routes
```
maebeam, line 1: merp

maebeam (approved):

PR #240, opened 2021-12-13
remove request body decoding and response body encoding from delete identities endpoint since both are empty objects

lazynina:


maebeam:
This endpoint can be removed

lazynina:
> This endpoint can be removed

PR #242, opened 2021-12-13
Give different sign-up bonus amounts based on country of ID

lazynina:
**Why?**

This change will allow us to adjust the amount users from each country receive when verifying with Jumio. Currently, the approach has been to turn off jumio for countries that have users who are trying to game the verification system. After this change, we'll be able to open up to more countries and modify the amount we give as a sign-up bonus.

**What's been done:**

- create utils directory and a constant that allows us to look up country names from country alpha-3 codes
- Add admin endpoints to get all sign-up bonus configurations and edit a single country's sign up bonus amount
- Default sign-up bonus amount is now specified in USD cents instead of DeSo nanos
- Update Jumio callback to use the new logic to get amount to pay referrer and referee
- Fix global state tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,7 @@ COPY backend/config  config
 COPY backend/cmd     cmd
 COPY backend/miner   miner
 COPY backend/routes  routes
+COPY backend/utils   utils
 COPY backend/main.go .
 
 # include core src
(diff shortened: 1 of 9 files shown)


triplegreenshell (changes requested):
Small refactoring change, otherwise looks good to me 👍

maebeam (approved):

PR #252, opened 2021-12-20
use min of richListLength constant and length of rich list when taking the top 1k

lazynina:
When running n0_test, taking the slice up to 1000 would result in a panic since the length of the rich list would be less than 1000


triplegreenshell (approved):
Looks good to me

PR #270, opened 2022-01-11
DAO Coin Endpoints

lazynina:


routes/transaction.go
```diff
@@ -2192,6 +2193,341 @@ func (fes *APIServer) SendDiamonds(ww http.ResponseWriter, req *http.Request) {
...
+	UpdaterPublicKeyBase58Check string `safeForLogging:"true"`
+
+	// The public key or username of the profile whose DAO coin the transactor is trying to transact with.
+	ProfilePublicKeyBase58CheckOrUsername string `safeForLogging:"true"`
+
+	// Whether this is a "mint", "burn" or "disable_minting" transaction
+	OperationType DAOCoinOperationTypeString `safeForLogging:"true"`
+
+	// Coins
+	CoinsToMintNanos uint256.Int `safeForLogging:"true"`
```
diamondhands0, line 2226: Does... this work? @lazynina

diamondhands0: Like can json serialize uint256?

lazynina: Yeah it becomes a hex string!!

diamondhands0: Hmmm looks like... maybe it works if you encode the things as a uint256? Not sure how to solve this problem...

https://github.com/holiman/uint256/blob/46b8ad94199afbcc12f2ca653481891165d829d5/conversion.go#L470

diamondhands0 (approved):

PR #276, opened 2022-01-18
Z/threaded comments

superzordon:
This PR adds new parameters to the `get-single-post` route which enables nested comments to be returned from the route. It also splits out the getComments piece of the route into it's own helper function to facilitate the necessary recursion.

The 3 parameters added are:
`ThreadLevelLimit`: How many reply-levels deep of comments will be retrieved. If unset, will only retrieve the top-level replies. For example, if this is set to 2, the api will return all replies to the parent post, replies to those replies, and replies to those replies.
`ThreadLeafLimit`: How many child replies of a parent comment will be considered when returning a comment thread. Setting this to -1 will include all child replies. This limit does not affect the top-level replies to a post. If this were set to one, the API would return all replies to a post, the top reply to each of those replies, the top reply to each of those replies, etc. Reply ranking is done in the same way comments are currently ordered.
`LoadAuthorThread`: If the post contains a comment thread where all comments are created by the author, include that thread in the response. Often, when someone wants to create a long post, they will do so by creating a post, reply to that post with page 2, reply to that reply with page 3 of the response, etc. If this flag is set, this full thread would be returned by the API.

### DIFF
--- a/routes/post.go
+++ b/routes/post.go
@@ -1462,7 +1462,7 @@ func (fes *APIServer) GetSinglePostComments(
 	// Limit comments to leaf limit, if it's not the first reply level and the leaf limit isn't -1
 	var limitedComments []*PostEntryResponse
 
-	if requestData.ThreadLeafLimit == -1 || commentLevel == 0 {
+	if requestData.ThreadLeafLimit == -1 || commentLevel == 0 || int32(len(comments)) <= requestData.ThreadLeafLimit {
 		limitedComments = comments
 	} else {
 		limitedComments = comments[:requestData.ThreadLeafLimit]
--- a/routes/post.go
+++ b/routes/post.go
@@ -1459,7 +1459,16 @@ func (fes *APIServer) GetSinglePostComments(
 		}
 	}
 
-	postEntryResponse.Comments = comments
+	// Limit comments to leaf limit, if it's not the first reply level and the leaf limit isn't -1
+	var limitedComments []*PostEntryResponse
+
+	if requestData.ThreadLeafLimit == -1 || commentLevel == 0 {
+		limitedComments = comments
+	} else {
+		limitedComments = comments[:requestData.ThreadLeafLimit]
+	}
+
+	postEntryResponse.Comments = limitedComments
 	return comments, nil
 }
 
--- a/routes/post.go
+++ b/routes/post.go
@@ -1194,12 +1194,11 @@ func (fes *APIServer) GetSinglePost(ww http.ResponseWriter, req *http.Request) {
 	comments, err := fes.GetSinglePostComments(
 		utxoView,
 		postEntryResponse,
-		ww,
 		requestData,
+		postEntry.PosterPublicKey,
 		readerPublicKeyBytes,
 		blockedPublicKeys,
 		0,
-		requestData.LoadAuthorThread,
 		postEntryResponse.PosterPublicKeyBase58Check,
 		)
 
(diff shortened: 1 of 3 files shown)


routes/post.go
```diff
@@ -1203,13 +1191,153 @@ func (fes *APIServer) GetSinglePost(ww http.ResponseWriter, req *http.Request) {
...
+		if profileEntry == nil {
+			continue
+		} else {
+			pubKeyToProfileEntryResponseMap[lib.MakePkMapKey(pubKeyBytes)] =
+				fes._profileEntryToResponse(profileEntry, utxoView)
+		}
+	}
+
+	// If the profile that posted this post does not have a profile, return with error.
+	if pubKeyToProfileEntryResponseMap[lib.MakePkMapKey(posterPublicKeyBytes)] == nil {
```
lazynina, line 1329: let save the PkMapKey for poster up above. It's a relatively inexpensive operation, but we really don't need to keep creating it.

routes/post.go
```diff
@@ -1203,13 +1191,153 @@ func (fes *APIServer) GetSinglePost(ww http.ResponseWriter, req *http.Request) {
...
+		return nil, err
```
lazynina, line 1330: err will be nil. Return a relevant error with fmt.Errorf

routes/post.go
```diff
@@ -1301,21 +1428,50 @@ func (fes *APIServer) GetSinglePost(ww http.ResponseWriter, req *http.Request) {
 	// Slice the comments from the offset up to either the end of the slice or the offset + limit, whichever is smaller.
 	maxIdx := lib.MinUint32(commentEntryResponseLength, requestData.CommentOffset+requestData.CommentLimit)
 	var comments []*PostEntryResponse
-	if commentEntryResponseLength > requestData.CommentOffset {
+	if commentEntryResponseLength > requestData.CommentOffset && commentLevel == 0 {
 		comments = commentEntryResponseList[requestData.CommentOffset:maxIdx]
+	} else {
+		comments = commentEntryResponseList
+	}
```
lazynina, line 1435: Can you add a small comment about why the different treatment based on comment level please?

lazynina (changes requested):
Looks pretty good. I think we will also want to enforce some basic sanity checks on that recursion so we don't wreck the servers.

Can you add some sanity checks on thread level limit and thread leaf limit

routes/post.go
```diff
@@ -1203,13 +1191,153 @@ func (fes *APIServer) GetSinglePost(ww http.ResponseWriter, req *http.Request) {
...
+				return nil, err
+			}
+			// If the currentPoster is not blacklisted (removed everywhere) and is greylisted (removed from leaderboard)
+			// add them back to the filteredProfilePubKeyMap and note that the currentPoster is greylisted.
+			if currentPosterUserMetadata.RemoveFromLeaderboard && !currentPosterUserMetadata.RemoveEverywhere {
+				isCurrentPosterGreylisted = true
+				filteredProfilePubKeyMap[lib.MakePkMapKey(posterPublicKeyBytes)] = posterPublicKeyBytes
+			}
+		}
+	}
```
lazynina, line 1309:
```suggestion
        posterPKID := utxoView.GetPKIDForPublicKey(posterPublicKeyBytes)
	isCurrentPosterGreylisted := false
	if fes.IsUserGraylisted(posterPKID.PKID) {
		isCurrentPosterGreylisted = true
		filteredProfilePubKeyMap[lib.MakePkMapKey(posterPublicKeyBytes)] = posterPublicKeyBytes
	}
```

and I think we replicate this bit of logic above in the original endpoint function

lazynina: actually do we need to check graylist status of each commenter? I think this logic was specifically put in for the poster of the post hash hex from the request data. If we don't need it here, we should kill it here. but may as well move the cleaner graylisted logic to the earlier bit if we can.

superzordon: This is a good point - this is no longer needed. I've updated accordingly.

lazynina (approved):

Issue #279, opened 2022-01-19
Using NodeSource triggers RuleErrorTxnFeeBelowNetworkMinimum error


tijno:
Running latest backend image. Tested #275 with `NODE_SOURCE=6` in dev.env but this causes issue with diamonding:

> E0119 12:02:44.144302       1 shared.go:30] SubmitTransaction: Problem processing transaction: VerifyAndBroadcastTransaction: Problem validating txn: ValidateTransaction: Problem validating transaction: : RuleErrorTxnFeeBelowNetworkMinimum

Not tested other transactions.

cc @superzordon

superzordon:
Interesting, haven't seen that before in my dev env. Do you experience this only when the `node-source` flag is set?

lazynina:
We need to add the extra data before we construct the transaction

lazynina:
Or increase your min fee rate nanos per kb a little until it is fixed

PR #281, opened 2022-01-20
Update inputs when adding node source to extra data

superzordon:
Closes https://github.com/deso-protocol/backend/issues/279


maebeam (changes requested):
This is not a good idea, would prefer to change the method signature for all of these methods to add ExtraData param

PR #285, opened 2022-01-28
Cache reserve exchange rate, buy deso fee basis points, and jumio default values from global state

lazynina:
Also more efficiently check if poster is only graylisted (not blacklisted) in get single post


triplegreenshell (approved):
This looks good to me - great improvement!

PR #306, opened 2022-02-21
Enable proper decoding for extra data values

tholonious:
Values in the ExtraData field can be encoded in different ways. The decoding needs to match the encoding used for each field.
Before:
```
"ExtraData": {
                "DiamondLevel": "\u0002",
                "DiamondPostHash": "~\ufffd+v\u2c45\u0010p\ufffdt\u000e%\ufffd\ufffd\fd\ufffd(\ufffd&\ufffd\ufffd\ufffdsO\u0019\u04fa\u0282$"
            },
```

After:
```
"ExtraData": {
                "DiamondLevel": "1",
                "DiamondPostHash": "7eba2b76e2b1851070f2740e25c1c70c64a228e226f0dff5734f19d3baca8224"
            },
```

### DIFF
--- a/routes/extra_data_utils.go
+++ b/routes/extra_data_utils.go
@@ -44,11 +44,11 @@ var ExtraDataKeyToDecoders = map[string]ExtraDataDecoder{
 // in transaction may have special encoding. In such cases
 // we'll need specialized decoders too
 func GetExtraDataDecoder(txnType lib.TxnType, key string) ExtraDataDecoder {
-	if txnType == lib.TxnTypeSubmitPost {
-		return DecodeString
-	}
 	if decoder, exists := ExtraDataKeyToDecoders[key]; exists {
 		return decoder
+	}
+	if txnType == lib.TxnTypeSubmitPost {
+		return DecodeString
 	}
 	// Default, just return hex encoding for bytes
 	return DecodeHexString
--- a/routes/extra_data_utils.go
+++ b/routes/extra_data_utils.go
@@ -1,0 +1,38 @@
+package routes
+
+import (
+	"encoding/hex"
+	"github.com/deso-protocol/core/lib"
+	"strconv"
+)
+
+type DecoderFunc func([]byte) string
+
+type ExtraDataDecoder struct {
+	Key     string
+	Decoder DecoderFunc
+}
+
+// GetExtraDataDecoderFunc Values in ExtraData field
+// in transaction may have special encoding. In such cases
+// we'll need specialized decoders too
+func GetExtraDataDecoderFunc(key string) DecoderFunc {
+	var decoders = []ExtraDataDecoder{
+		{
+			lib.DiamondLevelKey,
+			ByteArrTo64BitInt,
+		},
+	}
+	for _, decoder := range decoders {
+		if decoder.Key == key {
+			return decoder.Decoder
+		}
+	}
+	// Default, just return hex encoding for bytes
+	return hex.EncodeToString
+}
+
+func ByteArrTo64BitInt(bytes []byte) string {
+	var decoded, _ = lib.Varint(bytes)
+	return strconv.FormatInt(decoded, 10)
+}
--- a/routes/extra_data_utils.go
+++ b/routes/extra_data_utils.go
@@ -2,37 +2,88 @@ package routes
 
 import (
 	"encoding/hex"
+	"fmt"
 	"github.com/deso-protocol/core/lib"
 	"strconv"
 )
 
-type DecoderFunc func([]byte) string
+type ExtraDataDecoder func([]byte, *lib.DeSoParams) string
 
-type ExtraDataDecoder struct {
-	Key     string
-	Decoder DecoderFunc
+var ExtraDataKeyToDecoders = map[string]ExtraDataDecoder{
+	lib.RepostedPostHash:  DecodePkToString,
+	lib.IsQuotedRepostKey: DecodeBoolString,
+
+	lib.USDCentsPerBitcoinKey:      Decode64BitUintString,
+	lib.MinNetworkFeeNanosPerKBKey: Decode64BitUintString,
+	lib.CreateProfileFeeNanosKey:   Decode64BitUintString,
+	lib.CreateNFTFeeNanosKey:       Decode64BitUintString,
+	lib.MaxCopiesPerNFTKey:         Decode64BitUintString,
+
+	lib.ForbiddenBlockSignaturePubKeyKey: DecodePkToString,
+
+	lib.DiamondLevelKey:    Decode64BitIntString,
+	lib.DiamondPostHashKey: DecodeHexString,
+
+	lib.DerivedPublicKey: DecodePkToString,
+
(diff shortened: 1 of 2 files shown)


routes/extra_data_utils.go
```diff
@@ -0,0 +1,38 @@
...
+// GetExtraDataDecoderFunc Values in ExtraData field
+// in transaction may have special encoding. In such cases
+// we'll need specialized decoders too
+func GetExtraDataDecoderFunc(key string) DecoderFunc {
+	var decoders = []ExtraDataDecoder{
+		{
+			lib.DiamondLevelKey,
+			ByteArrTo64BitInt,
+		},
+	}
```
lazynina, line 25: Maybe we create a map outside of this function, called ExtraDataKeyToDecoder map so we don't have to iterate over all the decoders below.

routes/extra_data_utils.go
```diff
@@ -0,0 +1,38 @@
...
+
+// GetExtraDataDecoderFunc Values in ExtraData field
+// in transaction may have special encoding. In such cases
+// we'll need specialized decoders too
+func GetExtraDataDecoderFunc(key string) DecoderFunc {
+	var decoders = []ExtraDataDecoder{
+		{
+			lib.DiamondLevelKey,
+			ByteArrTo64BitInt,
+		},
```
lazynina, line 24: We should add decoders for the other consensus keys. You can find them listed here: https://github.com/deso-protocol/core/blob/1b9d4d61499d66f3c3f2911496785aa6b95dd5d5/lib/constants.go#L846.

Also maybe we converted ExtraData from a map of string to string into a map of string to interface

lazynina (changes requested):

routes/extra_data_utils.go
tholonious: Yeah, ExtraData should ultimately be a map of string to interface. Some ExtraData values themselves are maps, so we'll need to support some hierarchy. Best to make this change in a separate PR

routes/extra_data_utils.go
```diff
@@ -0,0 +1,64 @@
...
+func GetExtraDataDecoder(key string) ExtraDataDecoder {
+	if decoder, exists := ExtraDataKeyToDecoder[key]; exists {
+		return decoder
+	}
+
+	// Default, just return hex encoding for bytes
+	return hex.EncodeToString
+}
+
+func Decode64BitIntString(bytes []byte) string {
```
lazynina, line 55: I think we need a decode for both uint64 (using `lib.Uvarint`) and int64 (using `lib.Varint`).

routes/extra_data_utils.go
```diff
@@ -0,0 +1,64 @@
...
+	lib.SenderMessagingGroupKeyName:    hex.EncodeToString,
+	lib.RecipientMessagingPublicKey:    hex.EncodeToString,
+	lib.RecipientMessagingGroupKeyName: hex.EncodeToString,
+
+	lib.DESORoyaltiesMapKey: DecodePubKeyToUint64MapString,
+	lib.CoinRoyaltiesMapKey: DecodePubKeyToUint64MapString,
+
+	lib.MessagesVersionString: Decode64BitIntString,
+
+	lib.NodeSourceMapKey: DecodePubKeyToUint64MapString,
```
lazynina, line 40: The value is a uint64, not a map of pub key to uint64

routes/extra_data_utils.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"github.com/deso-protocol/core/lib"
+	"strconv"
+)
+
+type ExtraDataDecoder func([]byte) string
+
+var ExtraDataKeyToDecoder = map[string]ExtraDataDecoder{
+	lib.RepostedPostHash:  hex.EncodeToString,
+	lib.IsQuotedRepostKey: Decode64BitIntString,
```
lazynina, line 14: IsQuotedRepostKey maps to a value that represents a boolean

routes/extra_data_utils.go
```diff
@@ -0,0 +1,64 @@
...
+	"encoding/hex"
+	"fmt"
+	"github.com/deso-protocol/core/lib"
+	"strconv"
+)
+
+type ExtraDataDecoder func([]byte) string
+
+var ExtraDataKeyToDecoder = map[string]ExtraDataDecoder{
+	lib.RepostedPostHash:  hex.EncodeToString,
```
lazynina, line 13: I think we want to use lib.PkToString instead of hex.EncodeToString for post hashes and public keys.

lazynina (changes requested):

routes/extra_data_utils.go
tholonious: Yeah, it's represented as a byte with a value of {0, 1}. The functionality here that decode the value works the same for any length integer up to a length of 8 bytes. It still gives us a decoded value of {0, 1}. I'll create a new function for a more appropriate name for it to avoid confusion though

routes/extra_data_utils.go
```diff
@@ -0,0 +1,92 @@
...
+// GetExtraDataDecoder Values in ExtraData field
+// in transaction may have special encoding. In such cases
+// we'll need specialized decoders too
+func GetExtraDataDecoder(txnType lib.TxnType, key string) ExtraDataDecoder {
+	if txnType == lib.TxnTypeSubmitPost {
+		return DecodeString
+	}
+	if decoder, exists := ExtraDataKeyToDecoders[key]; exists {
+		return decoder
+	}
```
lazynina, line 52:
```suggestion
	if decoder, exists := ExtraDataKeyToDecoders[key]; exists {
		return decoder
	}
	if txnType == lib.TxnTypeSubmitPost {
		return DecodeString
	}
```

routes/extra_data_utils.go
```diff
@@ -0,0 +1,92 @@
...
+
+	lib.DiamondLevelKey:    Decode64BitIntString,
+	lib.DiamondPostHashKey: DecodeHexString,
+
+	lib.DerivedPublicKey: DecodePkToString,
+
+	lib.MessagingPublicKey:             DecodePkToString,
+	lib.SenderMessagingPublicKey:       DecodePkToString,
+	lib.SenderMessagingGroupKeyName:    DecodeString,
+	lib.RecipientMessagingPublicKey:    DecodeHexString,
```
lazynina, line 32:
```suggestion
	lib.RecipientMessagingPublicKey:    DecodePkToString,
```

routes/extra_data_utils.go
```diff
@@ -0,0 +1,92 @@
...
+	"encoding/hex"
+	"fmt"
+	"github.com/deso-protocol/core/lib"
+	"strconv"
+)
+
+type ExtraDataDecoder func([]byte, *lib.DeSoParams) string
+
+var ExtraDataKeyToDecoders = map[string]ExtraDataDecoder{
+	lib.RepostedPostHash:  DecodePkToString,
```
lazynina, line 13:
```suggestion
	lib.RepostedPostHash:  DecodeHexString,
```

lazynina (approved):

PR #317, opened 2022-03-20
Resolve error where get-nft-collection-summary fails if SN 1 is burned

triplegreenshell:


lazynina (approved):

lazynina (changes requested):
Actually we should only continue trying serial numbers up until the post's NumNFTCopies instead of the global max number of NFT copies

lazynina (approved):

PR #319, opened 2022-04-01
DAO Coin Exchange API endpoints

iamsofonias:


routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,183 @@
...
+type GetDAOCoinLimitOrdersRequest struct {
+	DAOCoin1CreatorPublicKeyBase58CheckOrUsername string `safeForLogging:"true"`
+	DAOCoin2CreatorPublicKeyBase58CheckOrUsername string `safeForLogging:"true"`
+}
+
+type GetDAOCoinLimitOrdersResponse struct {
+	Orders []DAOCoinLimitOrder
+}
+
+type DAOCoinLimitOrder struct {
```
lazynina, line 22: We typically use the `EntryResponse` suffix to denote a differentiation between the core struct and the backend struct.

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,183 @@
...
+	TransactorPublicKeyBase58CheckOrUsername string `safeForLogging:"true"`
```
lazynina, line 23: do we ever actually return a username here? or is it only ever a base58 check pub key? If it's only ever a pub key, denote it as such. Same applies to the Buying / Selling attributes below.

I think it makes most sense to change these attributes to be Base58Check and include ProfileEntryResponses for each in order to provide username / other details needed.

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,183 @@
...
+	BuyingDAOCoinCreatorPublicKeyBase58CheckOrUsername  string `safeForLogging:"true"`
+	SellingDAOCoinCreatorPublicKeyBase58CheckOrUsername string `safeForLogging:"true"`
+
+	// One of these two should be populated
+	ScaledExchangeRateCoinsToSellPerCoinToBuy *uint256.Int `safeForLogging:"true"`
+	ExchangeRateCoinsToSellPerCoinToBuy       float64      `safeForLogging:"true"`
+
+	// One of these two should be populated
+	QuantityToBuyInBaseUnits *uint256.Int `safeForLogging:"true"`
+	QuantityToBuy            float64      `safeForLogging:"true"`
```
lazynina, line 34: Will a float64 actually hold this? I'm not sure...

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,183 @@
...
+
+type DAOCoinLimitOrder struct {
+	TransactorPublicKeyBase58CheckOrUsername string `safeForLogging:"true"`
+
+	BuyingDAOCoinCreatorPublicKeyBase58CheckOrUsername  string `safeForLogging:"true"`
+	SellingDAOCoinCreatorPublicKeyBase58CheckOrUsername string `safeForLogging:"true"`
+
+	// One of these two should be populated
+	ScaledExchangeRateCoinsToSellPerCoinToBuy *uint256.Int `safeForLogging:"true"`
+	ExchangeRateCoinsToSellPerCoinToBuy       float64      `safeForLogging:"true"`
```
lazynina, line 30: will a float64 be big enough?

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,183 @@
...
+		)
+		if err != nil {
+			_AddBadRequestError(
+				ww,
+				fmt.Sprint("GetDAOCoinLimitOrders: Invalid DAOCoin1CreatorPublicKeyBase58CheckOrUsername"),
+			)
+			return
+		}
+		coin1PKID = utxoView.GetPKIDForPublicKey(coin1ProfilePublicKeyBytes).PKID
+		coin1ProfilePublicBase58Check = lib.Base58CheckEncode(coin1ProfilePublicKeyBytes, false, fes.Params)
```
lazynina, line 87: can probably refactor this into a function since this is all duplicated below. It can be a function that's defined within this function if that makes some things easier.

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,183 @@
...
+			ScaledExchangeRateCoinsToSellPerCoinToBuy:           order.ScaledExchangeRateCoinsToSellPerCoinToBuy,
+			ExchangeRateCoinsToSellPerCoinToBuy: floatExchangeRateCoinsToSellPerCoinToBuy(
+				order.ScaledExchangeRateCoinsToSellPerCoinToBuy,
+			),
+			QuantityToBuyInBaseUnits: order.QuantityToBuyInBaseUnits,
+			QuantityToBuy:            floatQuantityToBuy(order.QuantityToBuyInBaseUnits),
+		})
+	}
+
+	_ = json.NewEncoder(ww).Encode(GetDAOCoinLimitOrdersResponse{Orders: response})
```
lazynina, line 154: we need to catch this error. 
```suggestion
	if err = json.NewEncoder(ww).Encode(res); err != nil {
		_AddBadRequestError(ww, fmt.Sprintf("GetDAOCoinLimitOrders: Problem encoding response as JSON: %v", err))
		return
	}
```

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("GetDAOCoinLimitOrders: Error getting limit orders: %v", err))
+		return
+	}
+
+	var response []DAOCoinLimitOrder
+
+	for _, order := range ordersBuyingCoin1 {
+		transactorPublicKey := utxoView.GetPublicKeyForPKID(order.TransactorPKID)
+		response = append(response, DAOCoinLimitOrder{
```
lazynina, line 124: it might be better to write a function `_daoCoinLimitOrderToResponse` to handle this.

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,183 @@
...
+	exponent := big.NewInt(128)
+	divisor := big.NewFloat(0).SetInt(base.Exp(base, exponent, nil))
+
+	scaledValueAsBigFloat := big.NewFloat(0).SetInt(scaledValue.ToBig())
+
+	quotientAsBigFloat := big.NewFloat(0).Quo(
+		scaledValueAsBigFloat,
+		divisor,
+	)
+	quotient, _ := quotientAsBigFloat.Float64()
```
lazynina, line 169: probably need to watch out for overflow here. Not sure a float64 is gonna do the job.

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,183 @@
...
+
+// Given a quantity q, this returns q / (NanosPerUnit) as float
+func floatQuantityToBuy(quantityInBaseUnits *uint256.Int) float64 {
+	quantityInBaseUnitsAsBigFloat := big.NewFloat(0).SetInt(quantityInBaseUnits.ToBig())
+	divisor := big.NewFloat(float64(lib.NanosPerUnit))
+	quotientAsBigFloat := big.NewFloat(0).Quo(
+		quantityInBaseUnitsAsBigFloat,
+		divisor,
+	)
+	quotient, _ := quotientAsBigFloat.Float64()
```
lazynina, line 181: overflow concerns

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,183 @@
...
+	quotientAsBigFloat := big.NewFloat(0).Quo(
+		scaledValueAsBigFloat,
+		divisor,
+	)
+	quotient, _ := quotientAsBigFloat.Float64()
+	return quotient
+}
+
+// Given a quantity q, this returns q / (NanosPerUnit) as float
+func floatQuantityToBuy(quantityInBaseUnits *uint256.Int) float64 {
```
lazynina, line 174: if the return value of this is only used for frontend consumption, maybe we consider return a string? Worried about precision w/ a float64

routes/transaction.go
```diff
@@ -2548,6 +2547,287 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+	BuyingDAOCoinCreatorPublicKeyBase58CheckOrUsername string `safeForLogging:"true"`
+
+	// The public key or profile username of the DAO coin being sold
+	SellingDAOCoinCreatorPublicKeyBase58CheckOrUsername string `safeForLogging:"true"`
+
+	// Caller has the option of populating one of these two fields, depending on if they want
+	// to submit this param as a float or want to scale it themselves first. Populating both should
+	// result in an error response
+	ScaledExchangeRateCoinsToSellPerCoinToBuy *uint256.Int `safeForLogging:"true"`
+	ExchangeRateCoinsToSellPerCoinToBuy       float64      `safeForLogging:"true"`
```
lazynina, line 2588: will a float64 be big enough?

routes/transaction.go
```diff
@@ -2548,6 +2547,287 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+			fmt.Sprintf("CreateOrCancelDAOCoinLimitOrder: TransactionFees specified in Request body are invalid: %v", err),
+		)
+		return
+	}
+
+	buyingCoinPublicKey := lib.ZeroPublicKey.ToBytes()
+	sellingCoinPublicKey := lib.ZeroPublicKey.ToBytes()
+
+	if requestData.BuyingDAOCoinCreatorPublicKeyBase58CheckOrUsername != "" {
+		buyingCoinPublicKey, _, err = fes.GetPubKeAndProfileEntryForUsernameOrPublicKeyBase58Check(
```
lazynina, line 2689: lol I think I screwed up the name of this `GetPubKe` function. Can you please fix it as well?

routes/transaction.go
```diff
@@ -2548,6 +2547,287 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+			requestData.SellingDAOCoinCreatorPublicKeyBase58CheckOrUsername,
+			utxoView,
+		)
+		if err != nil {
+			return
+		}
+	}
+
+	scaledExchangeRateCoinsToSellPerCoinToBuy := requestData.ScaledExchangeRateCoinsToSellPerCoinToBuy
+	if requestData.ExchangeRateCoinsToSellPerCoinToBuy != 0 {
```
lazynina, line 2709:
```suggestion
	if requestData.ExchangeRateCoinsToSellPerCoinToBuy != float64(0) {
```

routes/transaction.go
```diff
@@ -2548,6 +2547,287 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+				"BuyingDAOCoinCreatorPublicKeyBase58CheckOrUsername or SellingDAOCoinCreatorPublicKeyBase58CheckOrUsername "+
+				"or both"),
+		)
+		return
+	}
+
+	if !validateExchangeRateCoinsToSellPerCoinToBuyParam(requestData) {
+		_AddBadRequestError(
+			ww,
+			fmt.Sprint("CreateOrCancelDAOCoinLimitOrder: Must provide only one of the "+
```
lazynina, line 2634:
```suggestion
			fmt.Sprint("CreateOrCancelDAOCoinLimitOrder: Must provide exactly one of the "+
```

routes/transaction.go
```diff
@@ -2548,6 +2547,287 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+				"ScaledExchangeRateCoinsToSellPerCoinToBuy or ExchangeRateCoinsToSellPerCoinToBuy params"),
+		)
+		return
+	}
+
+	if !validateQuantityToBuyParam(requestData) {
+		_AddBadRequestError(
+			ww,
+			fmt.Sprint("CreateOrCancelDAOCoinLimitOrder: Must provide only one of the "+
```
lazynina, line 2643:
```suggestion
			fmt.Sprint("CreateOrCancelDAOCoinLimitOrder: Must provide exactly one of the "+
```

lazynina (changes requested):

routes/transaction.go
iamsofonias: Discussed this offline. Yeah, float64 has a wide enough range

routes/transaction.go
iamsofonias: Not an issue here. The compiler is smart enough to know it's a float64 type when we use the 0 literal. The code reads a bit cleaner as-is

routes/dao_coin_exchange.go
iamsofonias: Discussed offline. float64 has a wider range than uint256.Int

routes/dao_coin_exchange.go
iamsofonias: Discussed offline. float64 has a wider range than uint256.Int

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,197 @@
...
+
+	adapter := utxoView.GetDbAdapter()
+
+	ordersBuyingCoin1, err := adapter.GetAllDAOCoinLimitOrdersForThisDAOCoinPair(coin1PKID, coin2PKID)
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("GetDAOCoinLimitOrders: Error getting limit orders: %v", err))
+		return
+	}
+
+	ordersSellingCoin1, err := adapter.GetAllDAOCoinLimitOrdersForThisDAOCoinPair(coin2PKID, coin1PKID)
```
lazynina, line 107: I think we should use `utxoView._getAllDAOCoinLimitOrdersForThisDAOCoinPair` instead of the dbAdapter so we also return results from the mempool

routes/transaction.go
```diff
@@ -2547,6 +2547,287 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+	buyingCoinPublicKey := lib.ZeroPublicKey.ToBytes()
+	sellingCoinPublicKey := lib.ZeroPublicKey.ToBytes()
+
+	if requestData.BuyingDAOCoinCreatorPublicKeyBase58CheckOrUsername != "" {
+		buyingCoinPublicKey, _, err = fes.GetPubKeyAndProfileEntryForUsernameOrPublicKeyBase58Check(
+			requestData.BuyingDAOCoinCreatorPublicKeyBase58CheckOrUsername,
+			utxoView,
+		)
+		if err != nil {
+			return
```
lazynina, line 2694: should add some error code here before returning - otherwise user will get a 200 response with no constructed tx

routes/transaction.go
```diff
@@ -2547,6 +2547,287 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+		}
+	}
+
+	if requestData.SellingDAOCoinCreatorPublicKeyBase58CheckOrUsername != "" {
+		sellingCoinPublicKey, _, err = fes.GetPubKeyAndProfileEntryForUsernameOrPublicKeyBase58Check(
+			requestData.SellingDAOCoinCreatorPublicKeyBase58CheckOrUsername,
+			utxoView,
+		)
+		if err != nil {
+			return
```
lazynina, line 2704: add some error here before returning

routes/transaction.go
```diff
@@ -2547,6 +2547,287 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+		}
+	}
+
+	quantityToBuyInBaseUnits := requestData.QuantityToBuyInBaseUnits
+	if requestData.QuantityToBuy != 0 {
+		quantityToBuyInBaseUnits, err = scaleQuantityToBuyToBaseUnits(
+			requestData.QuantityToBuy,
+		)
+		if err != nil {
+			if err != nil {
```
lazynina, line 2728: remove double if err != nil, only need one

lazynina (changes requested):

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,254 @@
...
+			SellingDAOCoinCreatorPublicKeyBase58Check: coin1ProfilePublicBase58Check,
+			ExchangeRateCoinsToSellPerCoinToBuy: calculateFloatExchangeRate(
+				order.ScaledExchangeRateCoinsToSellPerCoinToBuy,
+			),
+			QuantityToFill: calculateQuantityToFillAsFloat(order.QuantityToFillInBaseUnits),
+
+			OperationType: operationType,
+
+			OrderID: order.OrderID.String(),
+		})
```
lazynina, line 158: Can we refactor this into a function called `DAOCoinLimitOrderToResponse`?

routes/transaction.go
```diff
@@ -2547,6 +2547,275 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+		_AddBadRequestError(
+			ww,
+			"CreateDAOCoinLimitOrder: must provide at least one of BuyingDAOCoinCreatorPublicKeyBase58CheckOrUsername "+
+				"or SellingDAOCoinCreatorPublicKeyBase58CheckOrUsername",
+		)
+		return
+	}
+
+	// We only do basic validation of exchange rate and quantity here. If both are valid, then we pass the scaled
+	// values and remaining the request params to the helper for further validation
```
lazynina, line 2602:
```suggestion
	// values and the remaining request params to the helper for further validation
```

routes/transaction.go
```diff
@@ -2547,6 +2547,275 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+	}
+	scaledExchangeRateCoinsToSellPerCoinToBuy, err := lib.CalculateScaledExchangeRate(
+		requestData.ExchangeRateCoinsToSellPerCoinToBuy,
+	)
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("CreateDAOCoinLimitOrder: %v", err))
+		return
+	}
+
+	if requestData.ExchangeRateCoinsToSellPerCoinToBuy <= 0 {
```
lazynina, line 2619:
```suggestion
	if requestData.QuantityToFill <= 0 {
```

routes/transaction.go
```diff
@@ -2547,6 +2547,275 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+	cancelOrderIDBigInt, success := big.NewInt(0).SetString(requestData.CancelOrderID, 16)
+	if !success {
+		_AddBadRequestError(
+			ww,
+			fmt.Sprintf("CancelDAOCoinLimitOrder: Invalid CancelOrderID request param: %v", requestData.CancelOrderID),
+		)
+		return
+	}
+
+	cancelOrderIDBlockHash := lib.BigintToHash(cancelOrderIDBigInt)
```
lazynina, line 2686: See how we parse `postHashToModify` in `SubmitPost` (around line 1307 in transaction.go) for how we parse a block hash hex from the request

routes/transaction.go
```diff
@@ -2547,6 +2547,275 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+
+	res, err := fes.validateAndCreateDAOCoinLimitOrderResponse(
+		requestData.TransactorPublicKeyBase58Check,
+		"",
+		"",
+		uint256.NewInt(), // use default value for exchange rate and quantity, since we don't need to them to cancel an order
+		uint256.NewInt(),
```
lazynina, line 2693: we may be able to pass nil once the transaction construction function changes and the metadata encoding is fixed.

lazynina: or the construction function will be smart and just exclude these values.

routes/dao_coin_exchange.go
```diff
@@ -0,0 +1,254 @@
...
+			requestData.DAOCoin2CreatorPublicKeyBase58CheckOrUsername,
+		)
+		if err != nil {
+			_AddBadRequestError(
+				ww,
+				fmt.Sprintf("GetDAOCoinLimitOrders: Invalid DAOCoin2CreatorPublicKeyBase58CheckOrUsername: %v", err),
+			)
+			return
+		}
+	}
```
lazynina, line 97:
```suggestion
	
	getCoinPKIDAndPubKey := func(creatorPubKeyOrUsername string) (*lib.PKID, string, error) {
		coinPKID := &lib.ZeroPKID
		coinPublicKeyBase58Check := ""
		if creatorPubKeyOrUsername != "" {
			coinPublicKeyBase58Check, coinPKID, err = fes.validateCreatorPublicKeyBase58CheckOrUsername(
				utxoView,
				requestData.DAOCoin1CreatorPublicKeyBase58CheckOrUsername,
			)
			if err != nil {
				return nil, "", nil
			}
		}
		return coinPKID, coinPublicKeyBase58Check, nil
	}
	
	coin1PKID, coin1ProfilePublicBase58Check, err := getCoinPKIDAndPubKey(requestData.DAOCoin1CreatorPublicKeyBase58CheckOrUsername)
	if err != nil {
		_AddBadRequestError(
			ww,
			fmt.Sprintf("GetDAOCoinLimitOrders: Invalid DAOCoin1CreatorPublicKeyBase58CheckOrUsername: %v", err),
		)
	}

	coin2PKID, coin2ProfilePublicBase58Check, err := getCoinPKIDAndPubKey(requestData.DAOCoin1CreatorPublicKeyBase58CheckOrUsername)
	if err != nil {
		_AddBadRequestError(
			ww,
			fmt.Sprintf("GetDAOCoinLimitOrders: Invalid DAOCoin2CreatorPublicKeyBase58CheckOrUsername: %v", err),
		)
	}
```

lazynina: refactor into a locally scoped function

lazynina (changes requested):

routes/transaction.go
iamsofonias: Yeah, these can be nill once MF's PR is merged

routes/dao_coin_exchange.go
iamsofonias: Tried making this change locally. But TBH, I think the code reads more clearly as-is given a locally scoped function adds another level of indirection & nesting without deduplicating code any further. If you don't have strong feelings about this, my preference is to leave it as is.

lazynina: Fine by me - don't feel strongly about this

routes/utils.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"encoding/hex"
+	"github.com/deso-protocol/core/lib"
+	"github.com/pkg/errors"
+)
+
+// decodeBlockHashFromHex Decodes a BlockHash given a valid hex encoding. If the input does not represent a valid
+// BlockHash, this returns the corresponding error
+func decodeBlockHashFromHex(hexEncoding string) (*lib.BlockHash, error) {
```
iamsofonias, line 11: Moving forward, we should be using this function everywhere we decode block hashes from hex strings. I'll have another PR that migrates existing uses to use this function.

iamsofonias:
This is using Matt Foley's core branch https://github.com/deso-protocol/core/pull/329/files to test the endpoints against latest transaction metadata behavior with optional fields. 

The E2E testing looks good, but I'll hold off on merging until the core PR is merged into main too.

routes/transaction.go
```diff
@@ -2547,6 +2547,277 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+		lib.TxnTypeDAOCoinTransfer,
+		transactorPublicKeyBytes,
+		transactionFees,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("specified transactionFees are invalid: %v", err)
+	}
+
+	buyingCoinPublicKey := lib.ZeroPublicKey.ToBytes()
+	sellingCoinPublicKey := lib.ZeroPublicKey.ToBytes()
```
lazynina, line 2760: should we initialize these to empty byte slices instead so we can exclude them from txn construction? Lib.NewPublicKey returns nil if the number of bytes isn't 33
```suggestion
	var buyingCoinPublicKey []byte
	var sellingCoinPublicKey []byte
```

routes/transaction.go
```diff
@@ -2547,6 +2547,277 @@ func (fes *APIServer) TransferDAOCoin(ww http.ResponseWriter, req *http.Request)
...
+	txn, totalInput, changeAmount, fees, err := fes.blockchain.CreateDAOCoinLimitOrderTxn(
+		transactorPublicKeyBytes,
+		&lib.DAOCoinLimitOrderMetadata{
+			BuyingDAOCoinCreatorPublicKey:             lib.NewPublicKey(buyingCoinPublicKey),
+			SellingDAOCoinCreatorPublicKey:            lib.NewPublicKey(sellingCoinPublicKey),
+			ScaledExchangeRateCoinsToSellPerCoinToBuy: scaledExchangeRateCoinsToSellPerCoinToBuy,
+			QuantityToFillInBaseUnits:                 quantityToFillInBaseUnits,
+			OperationType:                             operationType,
+			CancelOrderID:                             cancelOrderId,
+			BidderInputs:                              nil,
```
lazynina, line 2791:
```suggestion
```
it's initialized to nil so we can delete this line.

lazynina (changes requested):

routes/transaction.go
iamsofonias: Ah, that's a good point. We'll want the public keys to be nil for order cancels. But we'll also want them to default to ZeroPublicKey when creating an order that's buying or selling DESO. I'll do a small refactor to account for this.

lazynina (approved):

PR #323, opened 2022-04-13
add subreddit hot feed

diamondhands0:


routes/hot_feed.go
```diff
@@ -764,6 +1016,17 @@ func (fes *APIServer) AdminUpdateHotFeedAlgorithm(ww http.ResponseWriter, req *h
...
+	if requestData.InteractionCapTag > 0 {
+		err := fes.GlobalState.Put(
+			_GlobalStatePrefixForHotFeedTagInteractionCap,
+			lib.EncodeUint64(uint64(requestData.InteractionCapTag)),
+		)
+		if err != nil {
+			_AddInternalServerError(ww, fmt.Sprintf("AdminUpdateHotFeedAlgorithm: Error putting InteractionCapTag: %v", err))
+			return
+		}
+	}
```
lazynina, line 1028: probably worth refactoring - define a function to store a constant in global state

routes/post.go
```diff
@@ -2292,3 +2290,28 @@ func GetPostHashFromPostHashHex(postHashHex string) (*lib.BlockHash, error) {
...
+	}
+	// Get body text from body and split on whitespace characters.
+	bodyString := bodyJSONObj.Body
+	bodyWords := strings.Fields(bodyString)
+
+	var tags []string
+
+	// Search each word to see if it's an @ mention (starts w/ @ and is at least of length 2).
+	for _, word := range bodyWords {
+		if len(word) >= 2 && word[0:1] == "@" {
```
lazynina, line 2310: note: this doesn't ensure that it's actually a user with a profile. so we're cool catching ALL @'s?

lazynina (changes requested):

routes/hot_feed.go
```diff
@@ -775,6 +1038,29 @@ func (fes *APIServer) AdminUpdateHotFeedAlgorithm(ww http.ResponseWriter, req *h
...
+
+	if len(requestData.TxnTypeMultiplierMap) > 0 {
+		txnTypeMultiplierMapBuffer := bytes.NewBuffer([]byte{})
+		if err := gob.NewEncoder(txnTypeMultiplierMapBuffer).Encode(requestData.TxnTypeMultiplierMap); err != nil {
+			_AddBadRequestError(ww, fmt.Sprintf("AdminUpdateHotFeedAlgorithm: Problem encoding transaction multiplier map: %v", err))
+			return
+		}
+		if err := fes.GlobalState.Put(_GlobalStatePrefixHotFeedTxnTypeMultiplierBasisPoints, txnTypeMultiplierMapBuffer.Bytes()); err != nil {
```
lazynina, line 1058: does it make sense for us to just store this in the frontend cache we use so that the value is updated immediately + we don't have to hit global state everytime the goroutine runs?

lazynina: applies to all these constants

routes/hot_feed.go
```diff
@@ -328,14 +395,57 @@ func (fes *APIServer) UpdateHotFeedOrderedList(
...
+			PostHashHex:  hex.EncodeToString(postHash[:]),
+			HotnessScore: hotnessInfo.HotnessScore,
+		}
+		hotFeedOrderedList = append(hotFeedOrderedList, hotFeedEntry)
 	}
+	sort.Slice(hotFeedOrderedList, func(ii, jj int) bool {
+		return hotFeedOrderedList[ii].HotnessScore > hotFeedOrderedList[jj].HotnessScore
+	})
+	fes.HotFeedOrderedList = hotFeedOrderedList
+	fes.HotFeedPostHashToTagScoreMap = hotnessInfoMapTagFeed
```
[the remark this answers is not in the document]
superzordon: We do not, as hotnessInfoMapTagFeed is a map. It will be converted to a sorted list in the `SaveOrderedFeedForTags` function.

routes/hot_feed.go
```diff
@@ -467,6 +479,20 @@ func (fes *APIServer) PopulateHotnessInfoMap(
 				continue
 			}
 
+			// If the post has been edited, remove all tags associated with that post.
+			// This ensures that the categorization reflects the most recently edited version.
+			isEditPost, postHashEdited := CheckTxnForEditPost(txn)
```
lazynina, line 484: sorry one last idea here: hidden posts? we should probably exclude them.

routes/hot_feed.go
```diff
@@ -750,10 +788,21 @@ func (fes *APIServer) GetHotnessScoreInfoForTxn(
...
+
+	// Get previously applied multiplier for post, if post has been counted already for this user.
+	if prevMultiplier, exists := postInteractionMap[interactionKey]; exists {
+		// If the previously applied multiplier is greater, skip this transaction.
+		if prevMultiplier > multiplier {
+			return nil, nil, 0
+		}
+		postInteractionMap[interactionKey] = multiplier
+		// We want to count the difference of the new multiplier and the previously counted multiplier.
+		multiplier = multiplier - prevMultiplier
```
lazynina, line 803: we're taking the diff here because this is only for the incremental multiplier achieved by this transaction right?

routes/hot_feed.go
```diff
@@ -1068,6 +1105,17 @@ func (fes *APIServer) AdminUpdateHotFeedAlgorithm(ww http.ResponseWriter, req *h
 	}
 }
 
+func (fes *APIServer) AddHotFeedParamToGlobalState(prefix []byte, value int) error {
+	if value > 0 {
+		err := fes.GlobalState.Put(
+			prefix,
+			lib.EncodeUint64(uint64(value)),
+		)
+		return err
```
lazynina, line 1114: I think we can just return without saving the error here.

routes/post.go
```diff
@@ -2305,11 +2305,11 @@ func ParseTagsFromPost(postEntry *lib.PostEntry) ([]string, error) {
 
 	var tags []string
 
-	// Search each word to see if it's an @ mention (starts w/ @ and is at least of length 2).
+	// Search each word to see if it's an @ mention, $ mention, or # tag (starts w/ symbol and is at least of length 2).
 	for _, word := range bodyWords {
-		if len(word) >= 2 && word[0:1] == "@" {
+		if len(word) >= 2 && (word[0:1] == "@" || word[0:1] == "#" || word[0:1] == "$") {
```
lazynina, line 2310: have we tested performance for catching these additional tags?

routes/hot_feed.go
```diff
@@ -410,59 +545,143 @@ func (fes *APIServer) UpdateHotFeedOrderedList(
...
+							newPostHashSet[*postHashScored] = true
+							fes.PostTagToPostHashesMap[tag] = newPostHashSet
+						} else {
+							// Only add to the set if the post hash isn't already present
+							if _, ok := postHashSet[*postHashScored]; !ok {
+								postHashSet[*postHashScored] = true
+							}
+							fes.PostTagToPostHashesMap[tag] = postHashSet
+						}
+					}
```
lazynina, line 576:
```suggestion
					for _, tag := range tags {
						// If a post hash set already exists, append to it,
						// otherwise create a new set and add it to the map.
						var postHashSet map[lib.BlockHash]bool
						if postHashSet, ok = fes.PostTagToPostHashesMap[tag]; !ok {
							postHashSet = make(map[lib.BlockHash]bool)
						}
						if _, ok = postHashSet[*postHashScored]; !ok {
							postHashSet[*postHashScored] = true
						}
						fes.PostTagToPostHashesMap[tag] = postHashSet
					}
```
fewer nested conditionals, easier for my pea brain to grasp

routes/hot_feed.go
```diff
@@ -328,26 +395,81 @@ func (fes *APIServer) UpdateHotFeedOrderedList(
...
+	postsToMultipliers map[lib.BlockHash]float64,
+	pkidsToMultipliers map[lib.PKID]*HotFeedPKIDMultiplier,
+	isTagFeed bool,
+) (map[lib.BlockHash]*HotnessPostInfo, error) {
 	hotnessInfoMap := make(map[lib.BlockHash]*HotnessPostInfo)
-	postInteractionMap := make(map[HotFeedInteractionKey][]byte)
+	// Map of interaction key to transaction type multiplier applied.
+	postInteractionMap := make(map[HotFeedInteractionKey]uint64)
 	for blockIdx, node := range relevantNodes {
 		block, _ := lib.GetBlock(node.Hash, utxoView.Handle)
```
lazynina, line 464: since we call PopulateHotnessInfoMap multiple times, we could optimize this by passing the relevant nodes in. GetBlock is a relatively expensive operation.

routes/post.go
triplegreenshell: Increases by about 50%, from ~1m to 1m30s

routes/hot_feed.go
triplegreenshell: Correct

lazynina:
I think once we have some cache busting mechanism, this should be good to go.

lazynina (approved):

PR #331, opened 2022-04-25
Use string <-> []byte casting as default for encoding extra data

iamsofonias:
The current hex.Decode inadvertently requires all input values to be valid hex encoded strings. We want to allow any string value, not just hex encoded ones.

### DIFF
--- a/routes/extra_data_utils.go
+++ b/routes/extra_data_utils.go
@@ -46,18 +46,13 @@ var ExtraDataKeyToDecoders = map[string]ExtraDataDecoder{
 	lib.TransactionSpendingLimitKey: DecodeTransactionSpendingLimit,
 }
 
-// GetExtraDataDecoder Values in ExtraData field
-// in transaction may have special encoding. In such cases
-// we'll need specialized decoders too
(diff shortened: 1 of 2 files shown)


routes/extra_data_utils.go
```diff
@@ -46,18 +46,13 @@ var ExtraDataKeyToDecoders = map[string]ExtraDataDecoder{
...
+// custom decoders. For all other cases, we use raw string <-> []byte casting for encoding & decoding.
+func GetExtraDataDecoder(_ lib.TxnType, key string) ExtraDataDecoder {
 	if decoder, exists := ExtraDataKeyToDecoders[key]; exists {
 		return decoder
 	}
-	if txnType == lib.TxnTypeSubmitPost {
-		return DecodeString
-	}
-	// Default, just return hex encoding for bytes
-	return DecodeHexString
```
iamsofonias, line 60: For context, this default was motivated by [LINK](https://github.com/deso-protocol/backend/pull/306), before we had a singular place to encode user-provided extra data values 

Now that we do, the default here should match the default used for encoding.

lazynina (approved):

PR #332, opened 2022-04-25
Allow phone number to be used multiple times up to a threshold

lazynina:


cmd/run.go
```diff
@@ -76,6 +76,10 @@ func init() {
 	runCmd.PersistentFlags().Uint64("min-satoshis-for-profile", 50000,
 		"Users won't be able to create a profile unless they buy this "+
 			"amount of satoshis or provide a phone number.")
+	// How many times can a phone number be used to receive starter DESO.
+	runCmd.PersistentFlags().Uint64("phone-number-use-threshold", 1, "A phone number will "+
```
diamondhands0, line 80: Let's set the default to 10 so we don't have to configure anything in infra. If it's too much we'll lower it with a flag.

lazynina, line 80:
```suggestion
	runCmd.PersistentFlags().Uint64("phone-number-use-threshold", 10, "A phone number will "+
```

diamondhands0:
Tested this locally and looks good.

@tijno I think your issue is distinct from this PR so let's cover it separately.

PR #336, opened 2022-04-27
Standardize handling of the ExtraData maps across backend API

iamsofonias:
This PR makes two changes:
1) It standardizes handling of client-defined ExtraData maps in all API endpoints when converting `map[string]string` <-> `map[string]byte`. String to binary casting is used by default to encode / decode field values.
2) It adds encoders in the API for the special ExtraData fields used in core. These special fields already have hard-coded encoding / decoding in core, which the API should match. If for any reason, a client wants to overwrite one of these extra data fields using the API, the API should be able to produce the same encoded values as core.

Change 2 is a nice-to-have that will reduce confusion long-term, but is not critical at this time. I'm open to dropping it from this PR.

In general, I think there should be two principles when working with ExtraData. For client-defined fields, the API should make as few assumptions as possible on how the clients want to encode the data (no more forced hex encoding), and all API endpoints should behave identically when accepting and returning ExtraData maps. For the special fields used in core, these should be treated as reserved fields with special behavior. The API should exactly match the encoding / decoding used in core.


routes/extra_data_encoding.go
```diff
@@ -0,0 +1,200 @@
...
+type ExtraDataDecoderFunc func([]byte, *lib.DeSoParams, *lib.UtxoView) string
+type ExtraDataEncoderFunc func(string) ([]byte, error)
+
+type ExtraDataEncoding struct {
+	Decode ExtraDataDecoderFunc
+	Encode ExtraDataEncoderFunc
+}
+
+// specialExtraDataKeysToEncoding These are reserved extra data fields used in core that have special encoding / decoding.
+// These fields are populated directly in core, but we still want to allow clients to be able to populate them directly
```
iamsofonias, line 22: Alternatively, we can use `ReservedFieldCannotOverride` to reject creation for any transaction where an ExtraData field populated by a client collides with one of these reserved fields used in core.

lazynina: I think it's okay to leave it flexible for now. The transactions will be rejected if the extra data has a reserved key that doesn't contain a legitimate value.

routes/extra_data_encoding.go
```diff
@@ -0,0 +1,200 @@
...
+		return DecodeString(bytes, params, utxoView)
+	}
+	response := TransactionSpendingLimitToResponse(transactionSpendingLimit, utxoView, params)
+	responseJSON, err := json.Marshal(response)
+	if err != nil {
+		glog.Errorf("Error marshaling transaction limit to string: %v", err)
+		// Fall back to default []byte -> string casting
+		return DecodeString(bytes, params, utxoView)
+	}
+	return string(responseJSON)
```
lazynina, line 199: What does this string look like?

routes/extra_data_encoding.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		glog.Errorf("Error marshaling public key to uint64 map to string: %v", err)
+		// Fall back to default []byte -> string casting
+		return DecodeString(bytes, params, utxoView)
+	}
+	mapWithDecodedKeys := map[string]uint64{}
+	for k, v := range decoded {
+		mapWithDecodedKeys[lib.PkToString(k.ToBytes(), params)] = v
+	}
+	return fmt.Sprint(mapWithDecodedKeys)
```
lazynina, line 182: What does this output look like?

routes/message.go
```diff
@@ -963,9 +972,14 @@ func (fes *APIServer) RegisterMessagingGroupKey(ww http.ResponseWriter, req *htt
 		return
 	}
 
+	extraData, err := EncodeExtraDataMap(requestData.ExtraData)
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("RegisterMessagingGroupKey: Problem encoding ExtraData: %v", err))
```
lazynina, line 977:
```suggestion
		_AddBadRequestError(ww, fmt.Sprintf("RegisterMessagingGroupKey: Problem encoding ExtraData: %v", err))
		return
```

routes/message.go
```diff
@@ -696,14 +700,19 @@ func (fes *APIServer) SendMessageStateless(ww http.ResponseWriter, req *http.Req
 		return
 	}
 
+	extraData, err := EncodeExtraDataMap(requestData.ExtraData)
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("SendMessageStateless: Problem encoding ExtraData: %v", err))
```
lazynina, line 705:
```suggestion
		_AddBadRequestError(ww, fmt.Sprintf("SendMessageStateless: Problem encoding ExtraData: %v", err))
		return
```

routes/transaction.go
```diff
@@ -2951,6 +2959,9 @@ func (fes *APIServer) createDAOCoinLimitOrderResponse(
 		return nil, err
 	}
 
+	txn.TxnMeta.(*lib.DAOCoinLimitOrderMetadata).FillType = 4
+	txn.TxnMeta.(*lib.DAOCoinLimitOrderMetadata).OperationType = 5
+
```
lazynina, line 2964:
```suggestion
```

This looks like leftover from when you were doing some testing. Please remove

lazynina (commented):
Overall, I think this is a good change to make at this time. I'd like @AeonSw4n and @diamondhands0 to weigh in on this as well since this will have some downstream impact.

routes/extra_data_encoding.go
iamsofonias: This decoder isn't changed by this PR. It only appears so because I renamed this file. I've changing the filename back to the original name so it's clearer what the PR changes

routes/extra_data_encoding.go
iamsofonias: Same as above. This implementation isn't changed

lazynina (approved):

PR #338, opened 2022-05-02
Conversion of DAO coin limit order exchange rates and quantities from float64 <-> uint256

iamsofonias:
The API converts exchange rates and quantities back and forth from the float64 values provided by clients to the uint256 values used in core. This PR adds proper handing for back and forth conversions between the two.

Exchange rate field:
- If an order is buying a DAO coin and selling $DESO, we scale down the exchange rate by (1e38 * 1e-9) to account for the fact that 1 $DESO is equal to 1e9 nanos whereas 1 DAO coin is equal to 1e18 base units
- If an order is buying $DESO and selling a DAO coin, then we scale the exchange rate by (1e38 * 1e9) for the same reason
- If an order is buying and selling $DAO coins, we scale by 1e38 only

The quantity field:
- If the quantity field refers to $DESO, then we use 1e9 for scaling float coin quantities to uint256 $DESO nanos
- If the quantity field refers to a DAO coin, we use 1e18 for scaling float coin quantities to uint256 DAO coin base units

These rules ensure that the exchange rates and quantities that core deals are always in base units, and the API does the leg work of properly interpreting them into float values for clients.

### DIFF
--- a/routes/dao_coin_exchange.go
+++ b/routes/dao_coin_exchange.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/deso-protocol/core/lib"
+	"github.com/golang/glog"
 	"github.com/holiman/uint256"
 	"github.com/pkg/errors"
 	"io"
@@ -241,6 +242,10 @@ func (fes *APIServer) buildDAOCoinLimitOrderResponsesForTransactor(
 			order,
 		)
 		if err != nil {
+			glog.Errorf(
+				"buildDAOCoinLimitOrderResponsesForTransactor: Unable to build DAO coin limit order response for limit order with OrderID: %v",
+				order.OrderID,
+			)
 			continue
 		}
 
@@ -317,7 +322,7 @@ func CalculateScaledExchangeRate(
 	if err != nil {
 		return nil, err
 	}
-	if rawScaledExchangeRate.Eq(uint256.NewInt().SetUint64(0)) {
+	if rawScaledExchangeRate.IsZero() {
 		return nil, errors.Errorf("The float value %f is too small to produce a scaled exchange rate", exchangeRateCoinsToSellPerCoinToBuy)
 	}
 	if buyingCoinPublicKeyBase58CheckOrUsername == "" {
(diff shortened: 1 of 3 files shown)


mattfoley8 (approved):

routes/dao_coin_exchange.go
```diff
@@ -269,61 +262,223 @@ func buildDAOCoinLimitOrderResponse(
...
+func CalculateScaledExchangeRate(
+	buyingCoinPublicKeyBase58CheckOrUsername string,
+	sellingCoinPublicKeyBase58CheckOrUsername string,
+	exchangeRateCoinsToSellPerCoinToBuy float64,
+) (*uint256.Int, error) {
+	rawScaledExchangeRate, err := lib.CalculateScaledExchangeRateFromString(formatFloatAsString(exchangeRateCoinsToSellPerCoinToBuy))
+	if err != nil {
+		return nil, err
+	}
+	if rawScaledExchangeRate.Eq(uint256.NewInt().SetUint64(0)) {
```
lazynina, line 320:
```suggestion
	if rawScaledExchangeRate.IsZero() {
```

routes/dao_coin_exchange.go
```diff
@@ -269,61 +262,223 @@ func buildDAOCoinLimitOrderResponse(
...
+		product := uint256.NewInt()
+		overflow := product.MulOverflow(rawScaledExchangeRate, getDESOToDAOCoinBaseUnitsScalingFactor())
+		if overflow {
+			return nil, errors.Errorf("Overflow when convering %f to a scaled exchange rate", exchangeRateCoinsToSellPerCoinToBuy)
+		}
+		return product, nil
+	} else if sellingCoinPublicKeyBase58CheckOrUsername == "" {
+		// Selling coin is $DESO
+		quotient := uint256.NewInt().Div(rawScaledExchangeRate, getDESOToDAOCoinBaseUnitsScalingFactor())
+		if quotient.Eq(uint256.NewInt().SetUint64(0)) {
```
lazynina, line 334:
```suggestion
		if quotient.IsZero() {
```

lazynina (commented):
The logic is sound, just a couple nits

PR #347, opened 2022-05-16
Use 'DESO' instead of empty string for DAO coin limit orders

iamsofonias:


lazynina (approved):

lazynina:
this makes a lot of sense given that go will interpret a missing string value in a struct as an empty string

PR #362, opened 2022-05-26
Mf/validate transactor selling qty

mattfoley8:
This is a secondary validation. We have a primary validation in core that also confirms transactor selling quantity. But this one can provide more immediate feedback to the user.

### DIFF
--- a/routes/dao_coin_exchange.go
+++ b/routes/dao_coin_exchange.go
@@ -607,13 +607,7 @@ func (fes *APIServer) validateTransactorSellingCoinBalance(
 		return errors.Errorf("Error decoding transactor public key: %v", err)
 	}
 
-	// Get selling coin PKID from public key.
-	sellingCoinPKID, err := fes.getPKIDFromPublicKeyBase58Check(
-		utxoView, requestData.SellingDAOCoinCreatorPublicKeyBase58Check)
-	if err != nil {
-		return errors.Errorf("Invalid SellingDAOCoinCreatorPublicKeyBase58Check: %v", err)
-	}
-
+	sellingCoinPKID := &lib.ZeroPKID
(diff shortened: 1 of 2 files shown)


routes/dao_coin_exchange.go
```diff
@@ -581,3 +581,117 @@ func getNumDigits(val *big.Int) int {
...
+			sellingCoinPKID.Eq(order.SellingDAOCoinCreatorPKID) {
+			// Calculate selling quantity.
+			orderSellingBaseUnits, err := order.BaseUnitsToSellUint256()
+			if err != nil {
+				return errors.Errorf("Error calculating open order selling quantity: %v", err)
+			}
+
+			// Sum selling quantity.
+			totalSellingBaseUnits, err = lib.SafeUint256().Add(totalSellingBaseUnits, orderSellingBaseUnits)
+			if err != nil {
```
iamsofonias, line 684: May as well make the `transactorSellingBalanceBaseUnits.Lt(totalSellingBaseUnits)` check here and exiting early if true

routes/dao_coin_exchange.go
```diff
@@ -581,3 +581,117 @@ func getNumDigits(val *big.Int) int {
...
+				return errors.Errorf("Error adding open order selling quantity: %v", err)
+			}
```
iamsofonias, line 686: May as well make the `transactorSellingBalanceBaseUnits.Lt(totalSellingBaseUnits)` check here and exiting early if true

iamsofonias, line 686: May as well make the `transactorSellingBalanceBaseUnits.Lt(totalSellingBaseUnits)` check here and exiting early if true

iamsofonias (approved):

PR #365, opened 2022-05-31
Tgs/hot feed updates

triplegreenshell:
This pull request adds in hot feed transactions to the current hot feed scoring mechanism. 

This update also reduces the lookback period for the hotfeed and as a result increases the cadence at which the hot feed is updated. 

This PR also removes logic that limits hot feed calculation to whenever new blocks are produced, as presumably every run there will be new txns in the mempool.

This PR also allows the scoring of posts with no interactions. This allows tagged feeds to include new posts without interaction.

### DIFF
--- a/routes/hot_feed.go
+++ b/routes/hot_feed.go
@@ -525,7 +525,7 @@ func (fes *APIServer) PopulateHotnessInfoMap(
 		glog.Errorf("Error getting mempool transactions: %v", err)
 	} else if len(txnsFromMempoolOrderedByTime) > 0 {
 		hotnessInfoBlocks = append(hotnessInfoBlocks, &HotnessInfoBlock{
-			Block:    &lib.MsgDeSoBlock{
+			Block: &lib.MsgDeSoBlock{
 				Txns: txnsFromMempoolOrderedByTime,
 			},
 			BlockAge: mempoolBlockHeight,
--- a/routes/hot_feed.go
+++ b/routes/hot_feed.go
@@ -82,7 +82,7 @@ func (fes *APIServer) StartHotFeedRoutine() {
 	out:
 		for {
 			select {
-			case <-time.After(30 * time.Second):
+			case <-time.After(10 * time.Second):
 				resetCache := false
 				if cacheResetCounter >= ResetCachesIterationLimit {
 					resetCache = true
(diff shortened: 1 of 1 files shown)


routes/hot_feed.go
```diff
@@ -517,6 +502,36 @@ func (fes *APIServer) PopulateHotnessInfoMap(
...
+	if len(hotnessInfoBlocks) > 0 {
+		mempoolBlockHeight = hotnessInfoBlocks[0].BlockAge + 1
+	} else {
+		mempoolBlockHeight = 1
+	}
+
+	// Create new "block" for mempool txns, give it a block age of 1 greater than the current tip
+
+	// First get all MempoolTxns from mempool.
+	dbMempoolTxnsOrderedByTime, _, err := fes.backendServer.GetMempool().GetTransactionsOrderedByTimeAdded()
```
lazynina: can we remove the `db` prefix on this variable name please?

routes/hot_feed.go
```diff
@@ -517,6 +502,36 @@ func (fes *APIServer) PopulateHotnessInfoMap(
...
+	if err != nil {
+		glog.Errorf("Error getting mempool transactions: %v", err)
+	} else if len(txnsFromMempoolOrderedByTime) > 0 {
+		hotnessInfoBlocks = append(hotnessInfoBlocks, &HotnessInfoBlock{
+			Block: &lib.MsgDeSoBlock{
+				Txns: txnsFromMempoolOrderedByTime,
+			},
+			BlockAge: mempoolBlockHeight,
+		})
+	}
```
lazynina: I generally prefer to have my `if err != nil` immediately following the declaration of that error. can you refactor this chunk of code such that we catch the error in the next line of code following its declaration?

routes/hot_feed.go
```diff
@@ -517,6 +502,36 @@ func (fes *APIServer) PopulateHotnessInfoMap(
...
 	// Map of interaction key to transaction type multiplier applied.
 	postInteractionMap := make(map[HotFeedInteractionKey]uint64)
+
+	// Fake block height for mempool transactions that haven't been mined yet
+	var mempoolBlockHeight int
+	if len(hotnessInfoBlocks) > 0 {
+		mempoolBlockHeight = hotnessInfoBlocks[0].BlockAge + 1
+	} else {
+		mempoolBlockHeight = 1
+	}
```
lazynina:
```suggestion
	mempoolBlockHeight := fes.block.BlockTip().Height + 1
```

lazynina (approved):

PR #369, opened 2022-06-06
Extend follow feed to be configurable length or infinite

piratekev:
This PR adds 2 new flags, `follow-feed-length` and `follow-feed-infinite` that allow node operators to configure their follow feeds to go back any amount of days or be infinite. The current behavior is a follow feed that goes back 2 days.


brootle:
@piratekev why did you close this one?

PR #372, opened 2022-06-10
Simulate limit and market orders to produce order execution preview

iamsofonias:


lazynina (approved):
Please add some comments on some of the stuff around txn fees and the other complicated pieces, but the logic here is sound.

PR #374, opened 2022-06-23
WIP: Do not merge

DeSoDog:
High level code execution for metamask gringotts integration

### DIFF
--- a/routes/shared.go
+++ b/routes/shared.go
@@ -2,17 +2,24 @@ package routes
 
 import (
 	"bytes"
+	"context"
 	"encoding/gob"
 	"encoding/json"
 	"fmt"
-	"github.com/holiman/uint256"
+	"math/big"
 	"net/http"
 	"time"
 
+	"github.com/holiman/uint256"
+
 	"github.com/deso-protocol/core/lib"
+	"github.com/ethereum/go-ethereum/params"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 	"github.com/tyler-smith/go-bip39"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
 )
 
 func _AddBadRequestError(ww http.ResponseWriter, errorString string) {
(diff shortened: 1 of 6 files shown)


routes/shared.go
```diff
@@ -342,9 +349,92 @@ func (fes *APIServer) putUserMetadataInGlobalState(
...
+		return nil, err
+	}
+	balance, err := utxoView.GetDeSoBalanceNanosForPublicKey(publicKey)
+	if err != nil {
+		return nil, err
+	}
+	if balance > 0 {
+		err := errors.New("account already has Deso")
+		return nil, err
+	}
```
AeonSw4n, line 373: We probably need to add an entry to the global state that says you received the money in the past. Imagine you get the starter money and then continously send it away, dropping your balance to zero.

routes/shared.go
```diff
@@ -342,9 +349,92 @@ func (fes *APIServer) putUserMetadataInGlobalState(
...
+	ethBalanceFloat := new(big.Float).SetInt(ethBalance)
+	minFloat, _ := new(big.Float).SetString(".05")
+	remainder := new(big.Float).Sub(ethBalanceFloat, minFloat)
+	fmt.Printf(remainder.String())
+	// need to figure out how to do float compare? Seems oddly confusing
+	// Anyways, if(minFloat > remainder {return nil, not error("not enough funds")"]
+
+	//2. Validations passed, lets construct a deso send transaction from gringotts to the new user and add it to the mem pool.
+	//  I'm not sure where to access the gringotts private key so for now I have dummy data being passed in
+	gringotts, err := bip39.EntropyFromMnemonic("replace this with gringotts value")
```
AeonSw4n, line 386: you can call `fes.SendSeedDeso`

routes/shared.go
```diff
@@ -342,9 +349,92 @@ func (fes *APIServer) putUserMetadataInGlobalState(
...
+	if err != nil {
+		return nil, fmt.Errorf("SendSeedDeSo: Problem processing starter seed transaction: %v", err)
+	}
+	// 3. return transaction action hash and we should be good. 
+	return txn.Hash(), nil
+
+}
 func (fes *APIServer) SendSeedDeSo(recipientPkBytes []byte, amountNanos uint64, useBuyDeSoSeed bool) (txnHash *lib.BlockHash, _err error) {
 	fes.mtxSeedDeSo.Lock()
+
```
AeonSw4n, line 437: let's remove redundant whitespaces

routes/shared.go
```diff
@@ -342,9 +349,92 @@ func (fes *APIServer) putUserMetadataInGlobalState(
...
+	balance, err := utxoView.GetDeSoBalanceNanosForPublicKey(publicKey)
+	if err != nil {
+		return nil, err
+	}
+	if balance > 0 {
+		err := errors.New("account already has Deso")
+		return nil, err
+	}
+	client, err := ethclient.Dial("https://mainnet.infura.io")
+	account := common.HexToAddress("public eth Address (not sure how to map deso key to eth key in go")
```
AeonSw4n, line 375: check `_verifyEthPersonalSignature` in core

routes/shared.go
```diff
@@ -342,9 +349,92 @@ func (fes *APIServer) putUserMetadataInGlobalState(
 
 	return nil
 }
+func etherToWei(val *big.Int) *big.Int {
+	return new(big.Int).Mul(val, big.NewInt(params.Ether))
+}
+func (fes *APIServer) SendDesoMetamask(recipientPkBytes string) (txnHash *lib.BlockHash, _err error) {
```
AeonSw4n, line 355: you also need to add an endpoint to server.go pointing to this. Trace through backend endpoints to see what's the pattern.

routes/shared.go
AeonSw4n: oh aren't we checking DeSo balance here? And yeah we could either make a global state entry (probably simplest) or check if they executed a transaction in the past -- though getting that view would be kinda tricky. I don't think we have a ready-to-use index for this but could be wrong.

PR #376, opened 2022-06-29
Validate DAO coin order transfer restriction.

mattfoley8:
This is already validated in core, but this check will provide more immediate feedback to the user if there is a relevant error message.

### DIFF
--- a/routes/dao_coin_exchange.go
+++ b/routes/dao_coin_exchange.go
@@ -1,6 +1,7 @@
 package routes
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"github.com/deso-protocol/core/lib"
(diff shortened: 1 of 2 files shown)


lazynina (approved):

routes/dao_coin_exchange.go
```diff
@@ -929,6 +930,57 @@ func (fes *APIServer) validateTransactorSellingCoinBalance(
...
+	if profileEntry.DAOCoinEntry.TransferRestrictionStatus == lib.TransferRestrictionStatusProfileOwnerOnly &&
+		!bytes.Equal(transactorPublicKey, buyingCoinPublicKey) {
+		return errors.New("Buying this DAO coin is restricted to the creator of the DAO")
+	}
+
+	// Validate if transfer restriction status is MEMBERS ONLY.
+	if profileEntry.DAOCoinEntry.TransferRestrictionStatus == lib.TransferRestrictionStatusDAOMembersOnly {
+		// Retrieve transactor's DAO coin balance. Error if balance is zero.
+		balanceEntry, _, _ := utxoView.GetBalanceEntryForHODLerPubKeyAndCreatorPubKey(transactorPublicKey, buyingCoinPublicKey, true)
+		if balanceEntry == nil || balanceEntry.BalanceNanos.IsZero() {
```
lazynina, line 976: Do we need to check if the balance entry is deleted here?

iamsofonias (approved):

PR #377, opened 2022-06-30
Calculate a profile's DAO coin market price without simulating order submission

iamsofonias:


routes/dao_coin_exchange.go
```diff
@@ -346,6 +346,60 @@ func buildDAOCoinLimitOrderResponse(
...
+	if err != nil {
+		return "", err
+	}
+	if len(orders) == 0 {
+		// It's OK if there are no orders on the book that allow us to exchange the coin pair. We default the exchange
+		// rate to 0 in this case
+		return "0", nil
+	}
+
+	bestOrderSellingOneBaseUnitOfTheSellingCoin := orders[0]
```
mattfoley8, line 376: nit: You could track the best price instead of the best order. That would save you from having to make copies of the order entry in the loop.

routes/user.go
```diff
@@ -985,54 +984,33 @@ func (fes *APIServer) _profileEntryToResponse(profileEntry *lib.ProfileEntry, ut
...
-				// In this case, the matching order is selling DESO to purchase DAO coins
-				// so the exchange rate is what we want without the need for inversion.
-			}
-			// Now we multiply by 1e9 as mentioned above
-			exchangeRate = NewHighPrecFloat().Mul(exchangeRate, NewHighPrecFloat().SetInt64(int64(lib.NanosPerUnit)))
+		profilePKID := utxoView.GetPKIDForPublicKey(profileEntry.PublicKey)
+		decimalPriceString, err := GetBestAvailableExchangeRateCoinsToBuyPerCoinToSell(
+			fes,
+			utxoView,
+			lib.NewPKID(lib.ZeroBlockHash[:]), // buying DESO
```
mattfoley8, line 991: There's a `lib.ZeroPKID` constant.

routes/dao_coin_exchange.go
```diff
@@ -346,6 +346,56 @@ func buildDAOCoinLimitOrderResponse(
...
 
+// GetBestAvailableExchangeRateCoinsToBuyPerCoinToSell computes the best available decimal string exchange rate at which
+// the market is able to exchange one base unit of the selling coin pair for the buying coin. Since we are interested
+// in computing the best exchange rate for the selling coin, the denominator for the output will always be the selling coin.
+//   Example: given buying coin B, and selling coin S, an output exchange rate of "1.5" implies an exchange rate of
+//            (1.5 coin B) per (1 coin S).
+// This function can support any arbitrary coin pair, but is most useful for markets where one coin is always considered
+// the denominating coin (ex: DAO coin <> DESO). In such cases, this computes the best available ask price.
+func GetBestAvailableExchangeRateCoinsToBuyPerCoinToSell(
+	fes *APIServer,
```
lazynina, line 357:
```suggestion
func (fes *APIServer) GetBestAvailableExchangeRateCoinsToBuyPerCoinToSell(
```
We typically make these functions on the APIServer struct instead of pass the APIServer as a param

routes/dao_coin_exchange.go
```diff
@@ -346,6 +346,56 @@ func buildDAOCoinLimitOrderResponse(
...
+		return "", err
+	}
+	if len(orders) == 0 {
+		// It's OK if there are no orders on the book that allow us to exchange the coin pair. We default the exchange
+		// rate to 0 in this case
+		return "0", nil
+	}
+
+	bestExchangeRate := orders[0].ScaledExchangeRateCoinsToSellPerCoinToBuy
+	for _, order := range orders[1:] {
```
lazynina, line 377:
```suggestion
	bestExchangeRate := uint256.NewInt()
	for _, order := range orders {
```
This will panic if we only have 1 order. Instead we should initialize the best rate to 0 and iterate over the entire slice

routes/dao_coin_exchange.go
```diff
@@ -789,6 +839,16 @@ func formatFloatAsString(f float64) string {
...
 
+// Returns float64 representation of the input decimal string. If the input is not a valid decimal string, this will
+// always return 0
+func tryParseFloatFromDecimalString(s string) (float64, error) {
+	float, err := strconv.ParseFloat(s, 64)
+	if err != nil {
+		return 0, err
+	}
+	return float, nil
+}
```
lazynina, line 850: This function is a little unnecessary - ParseFloat will return 0 if there's an error as the first return value.

lazynina (changes requested):

routes/dao_coin_exchange.go
iamsofonias: This is safe. The lower bound for slice expressions is the slice length so it can support my use-case here. [The spec for expressions](https://go.dev/ref/spec#Slice_expressions) is an interesting read if you're curious. Admittedly though, the code will probably be confusing for anyone else reading it in the future, and isn't worth putting here. I've updated it to your suggestion.

PR #380, opened 2022-07-08
Updated access bytes and buy deso with derived key

AeonSw4n:


routes/eth.go
```diff
@@ -367,14 +375,113 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
...
+	recipientPublicKey := lib.Base58CheckEncode(requestData.Signer, false, fes.Params)
+	recipientBytePK := []byte(recipientPublicKey)
+	// get the public eth address of the user
+	recipientEthAddress, err := publicKeyToEthAddress(requestData.Signer)
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf(DEFAULT_ERROR, err))
+	}
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf(DEFAULT_ERROR, err))
+	}
```
lazynina, line 408:
```suggestion
	if err != nil {
		_AddBadRequestError(ww, fmt.Sprintf(DEFAULT_ERROR, err))
		return
	}
```
we need to return after hitting an error.

routes/eth.go
```diff
@@ -367,14 +375,113 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
...
+	ethBalance := strings.Split(infuraResponse.Result.(string), "x")[1]
+	numberStr := strings.Replace(ethBalance, "0x", "", -1)
+	ethBalanceInt, err := strconv.ParseInt(numberStr, 16, 64)
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf(DEFAULT_ERROR, err))
+		return
+	}
+	// To prevent bots we only allow accounts with .001 eth or greater to qualify
+	if ethBalanceInt < int64(1000000000000000) { // the 1000000000000000 is equal to .001
```
lazynina, line 439: should we make this value configurable with a flag?

routes/eth.go
```diff
@@ -367,14 +375,113 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
...
+	params := []interface{}{recipientEthAddress, "latest"}
+	infuraResponse, err := fes.ExecuteETHRPCRequest("eth_getBalance", params)
+	// infura did something funky when getting the user balance
+	if infuraResponse == nil || err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf(DEFAULT_ERROR, "Infura balance request"))
+		return
+	}
+	ethBalance := strings.Split(infuraResponse.Result.(string), "x")[1]
+	numberStr := strings.Replace(ethBalance, "0x", "", -1)
+	ethBalanceInt, err := strconv.ParseInt(numberStr, 16, 64)
```
lazynina, line 433: Is this the easiest way to convert a hex balance into an int?

routes/eth.go
```diff
@@ -367,14 +375,113 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
...
+	}
+	// To prevent bots we only allow accounts with .001 eth or greater to qualify
+	if ethBalanceInt < int64(1000000000000000) { // the 1000000000000000 is equal to .001
+		_AddBadRequestError(ww, fmt.Sprintf("MetamaskSignin: To be eligible for airdrop your account needs to have more than .001 eth"))
+		return
+	}
+	//Verify that they signed a signature from their account
+	verifyEthError := lib.VerifyEthPersonalSignature(requestData.Signer, requestData.Message, requestData.Signature)
+	if verifyEthError != nil {
+		glog.Info("eth error:", verifyEthError)
```
lazynina, line 446:
```suggestion
```

routes/eth.go
```diff
@@ -367,14 +375,113 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
...
+		_AddBadRequestError(ww, fmt.Sprintf(DEFAULT_ERROR, verifyEthError))
+		return
+	}
+	// Converting to public key failed
+	if err != nil {
+		glog.Info("eth error:", err)
+		_AddBadRequestError(ww, fmt.Sprintf(DEFAULT_ERROR, err))
+		return
+
+	}
```
lazynina, line 456:
```suggestion
```

routes/eth.go
```diff
@@ -367,14 +375,113 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
...
+	// add them to the received airdrop list
+	if err = fes.GlobalState.Put(GlobalStateKeyMetamaskAirdrop(recipientBytePK), []byte{1}); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf(DEFAULT_ERROR, "GlobalState update failed"))
+		return
+	}
+
+	addressToAirdrop, _, err := lib.Base58CheckDecode(recipientPublicKey)
```
lazynina, line 463: catch this error? and should probably do this before we update global state

routes/eth.go
```diff
@@ -422,3 +529,14 @@ func (fes *APIServer) GetETHTransactionByHash(hash string) (_tx *InfuraTx, _err
 	}
 	return response, nil
 }
+func publicKeyToEthAddress(address []byte) (str string, err error) {
+	addressPubKey, err := btcutil.NewAddressPubKey(address, &chaincfg.MainNetParams)
+	if err != nil {
+		panic(err)
```
lazynina, line 535: Please just return the error.
```suggestion
		return "", err
```

routes/eth.go
```diff
@@ -422,3 +529,14 @@ func (fes *APIServer) GetETHTransactionByHash(hash string) (_tx *InfuraTx, _err
...
+	}
+	hash := sha3.NewLegacyKeccak256()
+	hash.Write(addressPubKey.PubKey().SerializeUncompressed()[1:])
+	sum := hash.Sum(nil)
+	str = "0x" + hex.EncodeToString(sum[12:])
+	return str, err
```
lazynina, line 541:
```suggestion
	return str, ""
```

lazynina (changes requested):
Solid changes, just a few small changes requested.

routes/user.go
```diff
@@ -3318,6 +3319,56 @@ func (fes *APIServer) GetTransactionSpendingLimitHexString(ww http.ResponseWrite
...
 }
 
+type GetAccessBytesRequest struct {
+	DerivedPublicKeyBase58Check string
+	ExpirationBlock             uint64
+	TransactionSpendingLimit    TransactionSpendingLimitResponse
+}
+
+type GetAccessBytesResponse struct {
+	SpendingLimitHex string
```
lazynina, line 3329: we should probably keep this naming consistent - I think we call it `TransactionSpendingLimitHex` in other places, but not  a huge deal either way.

routes/eth.go
AeonSw4n: good point!

routes/eth.go
DeSoDog: Turns out the first step wasn't needed but the last 2 were required from what I came across. Update has been pushed

lazynina (approved):

PR #381, opened 2022-07-14
Gringotts mvp

DeSoDog:


config/config.go
```diff
@@ -150,7 +151,7 @@ func LoadConfig(coreConfig *coreCmd.Config) *Config {
 	// ETH address to send all ETH received from "Buy With ETH" purchases.
 	config.BuyDESOETHAddress = viper.GetString("buy-deso-eth-address")
 	// Project ID for Infura requests
-	config.InfuraProjectID = viper.GetString("infura-project-id")
```
AeonSw4n, line 153: Let's revert to this

routes/eth.go
```diff
@@ -367,14 +369,104 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
 	}
 }
 
+type MetamaskSignInRequest struct {
+	RecipientPublicKey  string
+	RecipientEthAddress string
```
AeonSw4n, line 374: I think we should get rid of this parameter and instead compute it from `RecipientPublicKey`. Otherwise someone could just get infinite money by spamming this endpoint with random ETH addresses

routes/eth.go
```diff
@@ -367,14 +369,104 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
...
+	// validate the user's eth balance
+	params := []interface{}{requestData.RecipientEthAddress, "latest"}
+	infuraResponse, err := fes.ExecuteETHRPCRequest("eth_getBalance", params)
+	if infuraResponse == nil || err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("something wrong with fetching eth balance", err))
+		return
+	}
+	ethBalance := strings.Split(infuraResponse.Result.(string), "x")[1]
+	ethBalanceInt, _ := strconv.ParseInt(ethBalance, 16, 64)
+	ethBalanceInt64 := ethBalanceInt / 10000000000000000
```
AeonSw4n, line 424: Divisions are scary. Maybe we should just stick with big numbers. So like instead of the ethBalanceInt64 < 100, we just do int64(1000000000000000000)

routes/eth.go
```diff
@@ -367,14 +369,104 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		// glog the error, send generic message back for badrequest
+		_AddBadRequestError(ww, fmt.Sprintf("MetamaskSignin: Issues calculating user balance", err))
+		return
+	}
+	if ethBalanceInt64 < 100 { // the 100 is equal to .01, long story short converting in GO is hard
+		glog.Info(requestData.RecipientEthAddress, " has ", ethBalanceInt64)
+		glog.Info("requried amount is 100")
+
+		_AddBadRequestError(ww, fmt.Sprintf("MetamaskSignin: To be eligible for airdrop your account needs to have more than .05 eth"))
```
AeonSw4n, line 435: Shouldn't we say 0.01 ETH here? Also maybe we should decrease the threshold to just 0.001 ETH? Feels more inclusive to new users.

routes/eth.go
```diff
@@ -367,14 +369,104 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
...
+		return
+	}
+
+	//Verify that they signed a signature from their account TODO 1 re-enable this check after Identity changes
+	verifyEthError := lib.VerifyEthPersonalSignature(requestData.Signer, requestData.Message, requestData.Signature)
```
AeonSw4n, line 440: I think we should get rid of requestData.Signer and instead get the signer from RecipientPublicKey.

routes/eth.go
```diff
@@ -367,14 +369,104 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
...
+	if verifyEthError != nil {
+		glog.Info("eth error:", verifyEthError)
+		_AddBadRequestError(ww, fmt.Sprintf("MetamaskSignin: Invalid signer signature match #{verifyEthError}", verifyEthError))
+		return
+	}
+
+	// send deso to the user
+	// question do I need to do anything else to hook this up to gringott? add an env var maybe?
+	addressToAirdrop, _, _ := lib.Base58CheckDecode(requestData.RecipientPublicKey)
```
AeonSw4n, line 449: Please add error handling here.

routes/eth.go
```diff
@@ -367,14 +369,104 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
...
 	projectId := fes.Config.InfuraProjectID
 	URL := fmt.Sprintf("https://mainnet.infura.io/v3/%v", projectId)
-	if fes.Params.NetworkType == lib.NetworkType_TESTNET {
-		URL = fmt.Sprintf("https://ropsten.infura.io/v3/%v", projectId)
-	}
-
+	//if fes.Params.NetworkType == lib.NetworkType_TESTNET { TODO renable this
+	//	URL = fmt.Sprintf("https://ropsten.infura.io/v3/%v", projectId)
+	//}
+	glog.Info("infura url :", URL)
```
AeonSw4n, line 469: Let's revert

routes/eth.go
```diff
@@ -422,3 +514,10 @@ func (fes *APIServer) GetETHTransactionByHash(hash string) (_tx *InfuraTx, _err
 	}
 	return response, nil
 }
+
+func TextAndHash(data []byte) ([]byte, string) {
+	msg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", len(data), string(data))
+	hasher := sha3.NewLegacyKeccak256()
+	hasher.Write([]byte(msg))
+	return hasher.Sum(nil), msg
+}
```
AeonSw4n, line 523: Maybe we use the lib.TextAndHash function here?

scripts/nodes/n0_test
```diff
@@ -34,7 +34,7 @@ rm /tmp/main.*.log
   --num-mining-threads=1 \
   --miner-public-keys=BC1YLg7Bk5sq9iNY17bAwoAYiChLYpmWEi6nY6q5gnA1UQV6xixHjfV \
   --block-producer-seed='essence camp ghost remove document vault ladder swim pupil index apart ring' \
-  --starter-deso-seed='road congress client market couple bid risk escape artwork rookie artwork food' \
+  --starter-deso-seed='trash lake sense uphold bag nasty style gesture pizza hurry issue escape' \
```
AeonSw4n, line 37: Thanks for sharing your seed :) Let's revert cause people probably have this saved up.

AeonSw4n (commented):
Looks good! Let's just make sure someone can't trick the public keys.

routes/eth.go
DeSoDog: Function isn't used so we can just remove it

routes/eth.go
DeSoDog: Hmm couldn't i also derive the RecipientPublicKey from the message and signature?

routes/eth.go
```diff
@@ -367,14 +375,121 @@ func (fes *APIServer) QueryETHRPC(ww http.ResponseWriter, req *http.Request) {
 	}
 }
 
+type MetamaskSignInRequest struct {
+	AmountNanos uint64
+	Signer      []byte
```
AeonSw4n, line 380: Maybe let's change `Signer` to `UserPublicKeyBase58Check` and pass it in base58check string instead of a byte array? Can we also add some comments here to explain the params?

routes/shared.go
```diff
@@ -345,6 +346,7 @@ func (fes *APIServer) putUserMetadataInGlobalState(
 
 func (fes *APIServer) SendSeedDeSo(recipientPkBytes []byte, amountNanos uint64, useBuyDeSoSeed bool) (txnHash *lib.BlockHash, _err error) {
 	fes.mtxSeedDeSo.Lock()
+
```
AeonSw4n, line 349: can we remove this whitespace?

routes/shared.go
```diff
@@ -391,6 +393,7 @@ func (fes *APIServer) SendSeedDeSo(recipientPkBytes []byte, amountNanos uint64,
 		if err != nil {
 			return nil, err
 		}
+
```
AeonSw4n, line 396: can we remove this whitespace?

routes/eth.go
DeSoDog: Is there a benefit to doing that? Having signer, message, and signature all as byte arrays would make the request feel more uniform. We would also need to convert it back down to a byte array we deriving the eth public address anyways.

AeonSw4n: It displays better and feels more intuitive to pass strings rather than number arrays. It does work fine but ideally, the `message` and the `signature` should be hex strings too.

AeonSw4n:
Coolio

PR #383, opened 2022-07-27
Changed global state update

DeSoDog:


AeonSw4n (approved):
Ahh good good, makes sense to move it

PR #384, opened 2022-07-29
Add hot-feed-media-required parameter

piratekev:
This PR adds a configuration parameter hot-feed-media-required that, when set to true, will exclude non-media posts from being inserted into the Hot Feed.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -31,6 +31,7 @@ type Config struct {
 
 	// Hot Feed
 	RunHotFeedRoutine bool
+	HotFeedMediaRequired bool
 
 	// Web Security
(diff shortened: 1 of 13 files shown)


lazynina (approved):

config/config.go
```diff
@@ -31,6 +31,7 @@ type Config struct {
 
 	// Hot Feed
 	RunHotFeedRoutine bool
+	HotFeedMediaRequired bool
```
lazynina, line 34:
```suggestion
	RunHotFeedRoutine      bool
	HotFeedMediaRequired bool
```
Go-fmt is going to want this aligned

PR #391, opened 2022-08-31
Add get single derived key endpoint and update testnet eth url

lazynina:


routes/user.go
```diff
@@ -3263,6 +3249,80 @@ func (fes *APIServer) GetUserDerivedKeys(ww http.ResponseWriter, req *http.Reque
...
+type GetSingleDerivedKeyResponse struct {
+	DerivedKey *UserDerivedKey
+}
+
+func (fes *APIServer) GetSingleDerivedKey(ww http.ResponseWriter, req *http.Request) {
+	vars := mux.Vars(req)
+
+	ownerPublicKeyBase58Check, ownerPublicKeyExists := vars["ownerPublicKeyBase58Check"]
+	if !ownerPublicKeyExists {
+		_AddBadRequestError(ww, fmt.Sprintf("GetSingleDerivedKey: ownerPublicKeyBase58Check required"))
```
DeSoDog, line 3261: Add a return statement here

routes/server.go
```diff
@@ -1044,6 +1045,13 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
 			fes.GetUserDerivedKeys,
 			PublicAccess,
 		},
+		{
+			"GetSingleDerivedKey",
+			[]string{"GET"},
+			RoutePathGetSingleDerivedKey + "/{ownerPublicKeyBase58Check:[0-9a-zA-Z]{54,55}}/{derivedPublicKeyBase58Check:[0-9a-zA-Z]{54,55}}",
```
DeSoDog, line 1051: What does the regex mean here? Is that a type of validation guard?

DeSoDog (changes requested):

routes/user.go
lazynina: good catch. thanks!

routes/server.go
lazynina: it just ensures that we get a valid public key at this route param

PR #392, opened 2022-09-01
Fix backend function params

AeonSw4n:


lazynina (commented):
We need to add support in extra_data_utils.go to parse the MessagingGroupOperationType

PR #393, opened 2022-09-01
Add TransactionIDBase58Check to submit transaction response

lazynina:


DeSoDog (approved):
Looks straight forward to me. What's the benefit of adding TransactionIDBase58Check?

lazynina:
> Looks straight forward to me. What's the benefit of adding TransactionIDBase58Check?

this is a request from a 3rd party developer. Just makes things easier for them I guess. We use TransactionIDBase58Check in the explorer.

PR #409, opened 2022-11-12
"get-posts-for-public-key" endpoint: Added "IncludeComments" flag to …

DeSoEngineer:
…allow to also fetch comment posts.

See: https://github.com/deso-protocol/dips/discussions/93#discussioncomment-4120746

⚠ Depends on twin PR (https://github.com/deso-protocol/core/pull/417) in core.

### DIFF
--- a/routes/admin_referrals.go
+++ b/routes/admin_referrals.go
@@ -952,7 +952,7 @@ func (fes *APIServer) AdminDownloadRefereeCSV(ww http.ResponseWriter, req *http.
 		// it is not critical.
 		refereePostsLen := int64(-1)
 		refereePostEntries, err := utxoView.GetPostsPaginatedForPublicKeyOrderedByTimestamp(
-			refereePKID[:], nil, 1000, false, false)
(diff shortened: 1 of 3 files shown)


donhardman:
Are you sure that this will work? DeSo stores comments outside of posts for public key 👀

DeSoEngineer:
It is not about comments for individual posts, but for comment type posts, the public key authored.

That is, still, all returned posts are from the same pubkey, but those without a parent post are no longer filtered with "IncludeComments" == true.

But your feedback raises the question if I named the flag good enough. I purposely did not call it "FetchComments", to exactly avoid such confusion.

donhardman:
https://github.com/deso-protocol/core/blob/main/lib/db_utils.go#L6058-L6109

DeSoEngineer:
Ah, interesting, did not know that! 

If your quote is true, no, my code won't work 😊 

Will watch the approval flow. If my code does not work, we should also thinking about "fixing" the code I extended to avoid confusion in the future 😬

PR #417, opened 2022-11-28
Mf/add association txn types

mattfoley8:


routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+		_AddInternalServerError(ww, "CreateUserAssociation: problem getting public key for the transactor")
+		return
+	}
+
+	// Parse TargetUserPublicKeyBytes from TargetUserPublicKeyBase58Check.
+	if requestData.TargetUserPublicKeyBase58Check == "" {
+		_AddBadRequestError(ww, "CreateUserAssociation: must provide a TargetUserPublicKeyBase58Check")
+		return
+	}
+	targetUserPublicKeyBytes, _, err := fes.GetPubKeyAndProfileEntryForUsernameOrPublicKeyBase58Check(
```
lazynina, line 173: It's probably more efficient to just parse the public key instead of also fetching the profile.
Let's use `GetPubKeyBytesFromBase58Check` here instead (also applies to other places where we only need public key bytes and not profile)

routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+	}
+
+	// Validate AssociationType.
+	if requestData.AssociationType == "" {
+		_AddBadRequestError(ww, "CreateUserAssociation: must provide an AssociationType")
+		return
+	}
+
+	// Validate AssociationValue.
+	if requestData.AssociationValue == "" {
```
lazynina, line 203: is this a restriction in core too?

routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+		TransactionHex:    hex.EncodeToString(txnBytes),
+		TxnHashHex:        txn.Hash().String(),
+	}
+	if err = json.NewEncoder(ww).Encode(res); err != nil {
+		_AddInternalServerError(ww, "CreateUserAssociation: problem encoding response as JSON")
+		return
+	}
+}
+
+func (fes *APIServer) DeleteUserAssociation(ww http.ResponseWriter, req *http.Request) {
```
lazynina, line 267: do we also what to support a DeleteByAttributes?

routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+	}
+
+	// Fetch AssociationEntry.
+	associationEntry, err := utxoView.GetUserAssociationByID(lib.NewBlockHash(associationIdBytes))
+	if err != nil {
+		_AddInternalServerError(ww, fmt.Sprintf("GetUserAssociationByID: problem retrieving association: %v", err))
+		return
+	}
+	if associationEntry == nil {
+		_AddBadRequestError(ww, "GetUserAssociationByID: association not found")
```
lazynina, line 392: Let's use a `AddNotFoundError` here so we get 404s

routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+		return
+	}
+
+	// Convert AssociationEntry to AssociationResponse.
+	response := UserAssociationResponse{
```
lazynina, line 397: can we package this into a `_userAssociationEntryToResponse` function?

routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+			}
+			lastSeenAssociationID = lib.NewBlockHash(lastSeenAssociationIdBytes)
+		}
+	} else if queryType == AssociationQueryTypeCount {
+		if requestData.LastSeenAssociationID != "" {
+			return nil, errors.New("unsupported LastSeenAssociationID param for count operation")
+		}
+	} else {
+		return nil, errors.New("invalid query type") // This can never happen.
+	}
```
lazynina, line 567: let's use a switch - case statement for these. Additionally, I think we should do this towards the top of the function as these validations are cheaper than getting PKIDs from public keys. We can also move the validation on sort descending up there as well.

routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+
+	// Validate SortDescending.
+	if queryType == AssociationQueryTypeCount && requestData.SortDescending {
+		return nil, errors.New("unsupported SortDescending param for count operation")
+	}
```
lazynina, line 572: refactoring - move simple validations to top of function, use switch-case and consolidate logic for each query type
```suggestion
	// Parse Limit and LastSeenAssociationID.
	var lastSeenAssociationID *lib.BlockHash
	switch queryType {
	case AssociationQueryTypeQuery:
		if requestData.Limit < 0 || requestData.Limit > MaxAssociationsPerQueryLimit {
			return nil, errors.New("invalid Limit provided")
		}
		if requestData.Limit == 0 {
			requestData.Limit = MaxAssociationsPerQueryLimit
		}
		if requestData.LastSeenAssociationID != "" {
			lastSeenAssociationIdBytes, err := hex.DecodeString(requestData.LastSeenAssociationID)
			if err != nil {
				return nil, errors.New("invalid LastSeenAssociationID provided")
			}
			lastSeenAssociationID = lib.NewBlockHash(lastSeenAssociationIdBytes)
		}
	case AssociationQueryTypeCount:
		if requestData.Limit != 0 {
			return nil, errors.New("unsupported Limit param for count operation")
		}
		if requestData.LastSeenAssociationID != "" {
			return nil, errors.New("unsupported LastSeenAssociationID param for count operation")
		}
		if requestData.SortDescending {
			return nil, errors.New("unsupported SortDescending param for count operation")
		}
	default:
		return nil, errors.New("invalid query type") // This can never happen.
	}

	getPKIDForPublicKeyBase58Check := func (publicKeyBase58Check string) (*lib.PKID, error) {
		if publicKeyBase58Check == "" {
			return nil, nil
		}
		return fes.getPKIDFromPublicKeyBase58Check(
			utxoView, requestData.TargetUserPublicKeyBase58Check,
		)
	}

	transactorPKID, err := getPKIDForPublicKeyBase58Check(requestData.TransactorPublicKeyBase58Check)
	if err != nil {
		return nil, errors.New("problem getting PKID for the target transactor")
	}

	targetUserPKID, err := getPKIDForPublicKeyBase58Check(requestData.TargetUserPublicKeyBase58Check)
	if err != nil {
		return nil, errors.New("problem getting PKID for the target user")
	}

	appPKID, err := getPKIDForPublicKeyBase58Check(requestData.AppPublicKeyBase58Check)
	if err != nil {
		return nil, errors.New("problem getting PKID for the app")
	}
```

routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+		requestData.TransactorPublicKeyBase58Check,
+		utxoView,
+	)
+	if err != nil {
+		_AddInternalServerError(ww, "CreatePostAssociation: problem getting public key for the transactor")
+		return
+	}
+
+	// Parse PostHashBytes from PostHashHex.
+	if requestData.PostHashHex == "" {
```
lazynina, line 621: we probably want to validate that the post for this post hash hex really exists.

routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+		return
+	}
+	postHashBytes, err := hex.DecodeString(requestData.PostHashHex)
+	if err != nil {
+		_AddBadRequestError(ww, "CreatePostAssociation: invalid PostHashHex provided")
+		return
+	}
+
+	// Parse AppPublicKeyBytes from AppPublicKeyBase58Check.
+	if requestData.AppPublicKeyBase58Check == "" {
```
lazynina, line 632: global public key?

routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+		AssociationValue:       []byte(requestData.AssociationValue),
+		AssociationValuePrefix: []byte(requestData.AssociationValuePrefix),
+		Limit:                  requestData.Limit,
+		LastSeenAssociationID:  lastSeenAssociationID,
+		SortDescending:         requestData.SortDescending,
+	}
+	return assocationQuery, nil
+}
+
+func (fes *APIServer) CreatePostAssociation(ww http.ResponseWriter, req *http.Request) {
```
lazynina, line 590: same things from user associations apply to post associations. reviewing for things specific to post associations now.

routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+		requestData.TransactorPublicKeyBase58Check,
+		utxoView,
+	)
+	if err != nil {
+		_AddInternalServerError(ww, "DeleteUserAssociation: problem getting public key for the transactor")
+		return
+	}
+
+	// Parse AssociationIDBytes from AssociationID (hex string).
+	if requestData.AssociationID == "" {
```
lazynina, line 298: should we validate that the association exists and belongs to this transactor?

routes/associations.go
```diff
@@ -0,0 +1,1036 @@
...
+	// Count association entries.
+	count, err := utxoView.CountUserAssociationsByAttributes(associationQuery)
+	if err != nil {
+		_AddInternalServerError(ww, fmt.Sprintf("CountUserAssociations: %v", err))
+		return
+	}
+
+	// JSON encode response.
+	response := AssociationsCountResponse{Count: count}
+	if err = json.NewEncoder(ww).Encode(response); err != nil {
```
lazynina, line 484: we may just want to encode count as the response - it is probably just a touch faster and we don't really need a JSON object for this. additionally, we can axe the response struct definition.
```suggestion
	if err = json.NewEncoder(ww).Encode(count); err != nil {
```

routes/server.go
```diff
@@ -1132,6 +1136,76 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
...
+		{
+			"DeleteUserAssociation",
+			[]string{"POST", "OPTIONS"},
+			RoutePathUserAssociations + "/delete",
+			fes.DeleteUserAssociation,
+			PublicAccess,
+		},
+		{
+			"GetUserAssociationByID",
+			[]string{"GET", "OPTIONS"},
```
lazynina, line 1155: no options for gets
```suggestion
			[]string{"GET"},
```

lazynina (changes requested):

config/config.go
```diff
@@ -208,14 +208,18 @@ func LoadConfig(coreConfig *coreCmd.Config) *Config {
...
+		if !ok {
+			panic(fmt.Sprintf("Error parsing metamask-airdrop-eth-minimum into bigint: %v", metamaskAirdropMinStr))
+		}
+		var overflow bool
+		config.MetamaskAirdropEthMinimum, overflow = uint256.FromBig(metamaskAirdropMinBigint)
+		if overflow {
+			panic(fmt.Sprintf("metamask-airdrop-eth-minimum value %v overflows uint256", metamaskAirdropMinStr))
+		}
+	} else {
+		config.MetamaskAirdropEthMinimum = uint256.NewInt()
```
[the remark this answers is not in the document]
mattfoley8: Without this env variable the test server in the unit tests can't start up. All other env variables are optional (or have smart defaults), as I don't pass any to the test server.

routes/associations.go
mattfoley8: Yes.

routes/associations.go
mattfoley8: IMO no. GetByID mirrors DeleteByID. GetByAttributes is to support querying. If required, a user could do a GetByAttributes query then a subsequent DeleteByID. No sense in adding that complexity to core or backend IMO.

routes/associations.go
mattfoley8: Those validations exists in core. IMO there's some easy validations we should duplicate here (like a string isn't empty) but otherwise these endpoints should just be light wrappers around core functions and we should trust the core validations. Wdyt?

routes/associations.go
mattfoley8: That feels a little wonky to me. IMO not saving that much on bandwidth. And I always default to returning a proper JSON object. If for nothing else, future-proofs it in case we add additional fields.

routes/associations.go
mattfoley8: We do this in core. LMK if I should duplicate.

lazynina (approved):

PR #418, opened 2022-11-28
Add query param to get-single-profile-picture for content type

superzordon:


routes/user.go
```diff
@@ -1173,6 +1173,12 @@ func (fes *APIServer) GetSingleProfilePicture(ww http.ResponseWriter, req *http.
 		return
 	}
 
+	// Check to see if the content type is specified in the request.
+	contentTypeQueryParam := req.URL.Query().Get("contentType")
+	if contentTypeQueryParam != "" {
+		contentType = contentTypeQueryParam
+	}
```
lazynina, line 1180: We probably need some validation on content type?

lazynina (changes requested):

PR #422, opened 2022-12-23
Blacklist and graylist by username

superzordon:
Currently, nodes are only able to blacklist by profile. This means that if an account is using a deceptive username that should be blocked, once blacklisted it can simply switch the name to a new profile and continue on. This change creates a list of blocked usernames that will always be blocked, rather than tracking the profiles themselves.

### DIFF
--- a/routes/admin_user.go
+++ b/routes/admin_user.go
@@ -231,7 +231,7 @@ type AdminUpdateUsernameBlacklistRequest struct {
 // AdminUpdateUsernameBlacklist...
 //
 // This endpoint allows an admin to blacklist or graylist an account by username.
-// This is important as it prevents malicous users using a deceiptful username from continuing to username change
+// This is important as it prevents malicious users using a deceitful username from continuing to username change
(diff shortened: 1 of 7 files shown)


routes/admin_user.go
```diff
@@ -214,6 +214,87 @@ func (fes *APIServer) AdminUpdateUserGlobalMetadata(ww http.ResponseWriter, req
...
+	// Set to true if this user's content should not show up anywhere on the site.
+	AddUserToList bool `safeForLogging:"true"`
+
+	AdminPublicKey string
+}
+
+// AdminUpdateUsernameBlacklist...
+//
+// This endpoint allows an admin to blacklist or graylist an account by username.
+// This is important as it prevents malicous users using a deceiptful username from continuing to username change
```
lazynina, line 234:
```suggestion
// This is important as it prevents malicious users using a deceitful username from continuing to username change
```

lazynina (approved):

PR #425, opened 2023-01-09
add AssociationLimitMap to TransactionSpendingLimitResponse

lazynina:


routes/transaction.go
```diff
@@ -3186,6 +3195,11 @@ type TransactionSpendingLimitResponse struct {
 	// of SellingCoinPublicKey mapped to the number of DAO Coin Limit Order transactions with
 	// this Buying and Selling coin pair that the derived key is authorized to perform.
 	DAOCoinLimitOrderLimitMap map[string]map[string]uint64
+	// AssociationLimitMap is a map with AssociationClass { User | Post } as keys mapped to a map of
```
mattfoley8, line 3198: This comment needs to be updated w/ the new data struct.

routes/transaction.go
```diff
@@ -3471,6 +3485,29 @@ func TransactionSpendingLimitToResponse(
...
 	}
+
+	if len(transactionSpendingLimit.AssociationLimitMap) > 0 {
+		for associationLimitKey, opCount := range transactionSpendingLimit.AssociationLimitMap {
+			associationClassString := associationLimitKey.AssociationClass.ToAssociationClassString()
+			associationType := associationLimitKey.AssociationType
+			associationAppScopeTypeString := associationLimitKey.AppScopeType.ToAssociationAppScopeTypeString()
+			associationOperationString := associationLimitKey.Operation.ToAssociationOperationString()
+			var appPublicKey string
+			if !associationLimitKey.AppPKID.IsZeroPKID() {
```
mattfoley8, line 3496: If the AppPKID is the ZeroPKID, don't we want to set the appPublicKey to the ZeroPublicKey? Or you're thinking we'll leave it as the empty string?

routes/transaction.go
```diff
@@ -3580,6 +3617,26 @@ func (fes *APIServer) TransactionSpendingLimitFromResponse(
 		}
 	}
 
+	if len(transactionSpendingLimitResponse.AssociationLimitMap) > 0 {
+		transactionSpendingLimit.AssociationLimitMap = make(map[lib.AssociationLimitKey]uint64)
+		for _, associationLimitMapItem := range transactionSpendingLimitResponse.AssociationLimitMap {
+			appPKID := &lib.ZeroPKID
```
mattfoley8, line 3623: Here, you default to the ZeroPKID. We should probably do the same above and default to the ZeroPublicKey.

lazynina: we need the PKID for the TransactionSpendingLimit Struct. TransactionSpendingLimitResponse gives us a public key and we need to convert it to a PKID for TransactionSpendingLimit. When we're going from TransactionSpendingLimit to TransactionSpendingLImitResponse, we convert the PKID to a public key

routes/transaction.go
lazynina: I guess this depends on the `AppScopeType`, right? so we want to differentiate between the global app public key and the "any" app public key (which would be represented by the empty string)

PR #426, opened 2023-01-12
Add count associations by each association value endpoint.

mattfoley8:


routes/associations.go
```diff
@@ -124,6 +145,10 @@ type AssociationsCountResponse struct {
 	Count uint64
 }
 
+type AssociationCountsReponse struct {
+	Counts map[string]uint64
```
lazynina, line 149: Let's add a total field to make it a little easier for consumers of this API

lazynina (changes requested):
Let's just add the total to the response, but otherwise this looks good.

lazynina (approved):

PR #432, opened 2023-01-18
Get snapshot epoch metadata backend endpoint

AeonSw4n:


routes/server.go
```diff
@@ -295,6 +295,10 @@ const (
 	// associations.go
 	RoutePathUserAssociations = "/api/v0/user-associations"
 	RoutePathPostAssociations = "/api/v0/post-associations"
+
+	// snapshot.go
+	RoutePathGetSnapshotEpochMetadata = "/api/v0/get-snapshot-epoch-metadata"
+	RoutePathGetStateChecksum         = "/api/v0/get-state-checksum"
```
lazynina, line 301: since these are really `GET` routes, we don't need to include `get` in the route.

routes/server.go
```diff
@@ -2026,6 +2045,121 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
 			fes.GetAllUserMessageThreads,
 			PublicAccess,
 		},
+		// registering the routes related to access groups
+		{
```
lazynina, line 2049: shouldn't all these routes already be in there?

lazynina (changes requested):

lazynina (approved):

lazynina (approved):

PR #473, opened 2023-04-18
update GetTransactionSpending for balance model

lazynina:


routes/transaction.go
```diff
@@ -3899,11 +3905,17 @@ func (fes *APIServer) GetTransactionSpending(ww http.ResponseWriter, req *http.R
 	}
 
 	// Return the final transaction spending.
-	totalSpendingNanos := totalInputNanos - changeAmountNanos
+	var totalSpendingNanos uint64
+	if totalInputNanos > changeAmountNanos {
```
mattfoley8, line 3909: nit: should we include an error case else case?

mattfoley8 (approved):
One small nit. Feel free to address or disregard. LGTM.

PR #501, opened 2023-06-14
Mf/add get validator by public key endpoint

mattfoley8:


routes/server.go
```diff
@@ -1287,6 +1287,13 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
 			fes.UnregisterAsValidator,
 			PublicAccess,
 		},
+		{
+			"GetValidatorByPublicKeyBase58Check",
+			[]string{"GET", "OPTIONS"},
```
lazynina, line 1292:
```suggestion
			[]string{"GET"},
```
Do we need options here?

routes/server.go
```diff
@@ -1287,6 +1287,13 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
...
+			RoutePathValidators + "/{publicKeyBase58Check:[0-9a-zA-Z]{54,55}}",
```
lazynina, line 1293: We could probably use a better base58 check regex. Here are the regexes we're using in deso-js to verify that the pub key is of a valid form:
```
const regexMainnet = /^BC[1-9A-HJ-NP-Za-km-z]{53}$/;
const regexTestnet = /^tBC[1-9A-HJ-NP-Za-km-z]{51}$/;
```

routes/validators.go
```diff
@@ -210,3 +225,64 @@ func (fes *APIServer) UnregisterAsValidator(ww http.ResponseWriter, req *http.Re
...
+	}
+
+	// Convert Domains [][]byte to []string.
+	var domains []string
+	for _, domain := range validatorEntry.Domains {
+		domains = append(domains, string(domain))
+	}
+
+	// Convert ValidatorEntry to ValidatorResponse.
+	validatorResponse := &ValidatorResponse{
```
lazynina, line 270: let's move the conversion of the core ValidatorEntry type to the backend ValidatorResponse type to a separate function in case we need to re-use it.

lazynina (changes requested):

lazynina (approved):

PR #506, opened 2023-08-11
Update badger sync settings to optimize memory usage during hypersync

superzordon:
Update hypersync to use default badger settings, switch to performance settings once hypersync completes


diamondhands0 (approved):
This looked good

PR #509, opened 2023-08-27
Add captcha verification

superzordon:


routes/global_state.go
```diff
@@ -673,6 +682,12 @@ func GlobalStateKeyForBlacklistedProfileByUsername(username string) []byte {
 	return key
 }
 
+// Key for accessing a blacklisted user by username.
```
lazynina, line 685: Let's update this comment.

routes/verify.go
```diff
@@ -251,6 +252,208 @@ func (fes *APIServer) validatePhoneNumberNotAlreadyInUse(phoneNumber string, use
...
+	userMetadata, err := fes.getUserMetadataFromGlobalState(publicKeyBase58Check)
+	if err != nil {
+		return "", fmt.Errorf("HandleCaptchaVerificationRequest: Problem with getUserMetadataFromGlobalState: %v", err)
+	}
+
+	if userMetadata.LastHcaptchaBlockHeight != 0 {
+		return "", fmt.Errorf("HandleCaptchaVerificationRequest: LastHcaptchaBlockHeight is already set")
+	}
+
+	if fes.Config.StarterDESOSeed == "" {
```
lazynina, line 392: we should move this check higher up since we don't need to verify the captcha or hit global state if we're in this state.

routes/verify.go
```diff
@@ -251,6 +252,208 @@ func (fes *APIServer) validatePhoneNumberNotAlreadyInUse(phoneNumber string, use
...
+		return "", fmt.Errorf("HandleCaptchaVerificationRequest: Starter DESO seed not set")
+	}
+
+	// Retrieve the amount of nanos to reward for a successful captcha.
+	amountToSendNanos, err := fes.getCaptchaRewardNanosFromGlobalState()
```
lazynina, line 397: we can also do this one earlier. If the amount to send is 0, we can return early.

routes/verify.go
```diff
@@ -251,6 +252,208 @@ func (fes *APIServer) validatePhoneNumberNotAlreadyInUse(phoneNumber string, use
...
+	userMetadata.LastHcaptchaBlockHeight = lastBlockheight
+	userMetadata.HcaptchaShouldCompProfileCreation = true
+
+	if err = fes.putUserMetadataInGlobalState(userMetadata); err != nil {
+		return "", fmt.Errorf("HandleCaptchaVerificationRequest: Problem with putUserMetadataInGlobalState: %v", err)
+	}
+
+	// Send the starter DESO to the user.
+	var txnHash *lib.BlockHash
+	publicKeyBytes, _, err := lib.Base58CheckDecode(publicKeyBase58Check)
```
lazynina, line 413: Handle this error? also, maybe do this earlier in the function (or parse it in the calling function)? We probably want to have all inputs verified before we call this function.

lazynina (changes requested):

lazynina (approved):

PR #539, opened 2024-02-08
Fix empty verification map on error

erwin-willems:
A few times a month, our Verified Users data disappear. I think this is because of a retrieval error from badgerdb. This fix is to make sure the Verified Users are not overwritten with an empty map.

I think error handling on txn.Get needs to be reviewed

### DIFF
--- a/routes/shared.go
+++ b/routes/shared.go
@@ -260,18 +260,18 @@ func (fes *APIServer) GetVerifiedUsernameToPKIDMapFromGlobalState() (_verificati
 		}
 	} else {
 		// Create the inital map structure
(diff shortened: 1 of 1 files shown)


routes/global_state.go
```diff
@@ -980,6 +980,8 @@ func (gs *GlobalState) Get(key []byte) (value []byte, _err error) {
 	err := gs.GlobalStateDB.View(func(txn *badger.Txn) error {
 		item, err := txn.Get(key)
 		if err != nil {
+			// Todo: I think this error needs to be processed. It could be a key not found error, then nil should be returned.
```
lazynina, line 983: I think handling the error here means we do not need to delete the code below. Use errors.Is to compare err to badger's key not found error.

lazynina (changes requested):

PR #565, opened 2024-03-25
Fix minor issues to speed up block processing logic

diamondhands0:
* The Dockerfile change in this PR switches us to compiling with debug
  settings. This slows things down a bit but it's worth it because it
means we can attach debugging instances directly to a running node
remotely.

* Otherwise, this is just extra logging.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,14 +30,31 @@ COPY core/cmd       ../core/cmd
 COPY core/lib       ../core/lib
 COPY core/migrate   ../core/migrate
 
+# Install Delve debugger, specifying the installation path explicitly
+ENV GOPATH=/root/go
+RUN go install github.com/go-delve/delve/cmd/dlv@latest
+
 # build backend
-RUN GOOS=linux go build -mod=mod -a -installsuffix cgo -o bin/backend main.go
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -30,14 +30,27 @@ COPY core/cmd       ../core/cmd
...
 COPY core/migrate   ../core/migrate
 
+# Install Delve debugger, specifying the installation path explicitly
+ENV GOPATH=/root/go
+RUN go install github.com/go-delve/delve/cmd/dlv@latest
+
 # build backend
-RUN GOOS=linux go build -mod=mod -a -installsuffix cgo -o bin/backend main.go
+#RUN GOOS=linux go build -mod=mod -a -installsuffix cgo -o bin/backend main.go
+RUN GOOS=linux go build -gcflags="all=-N -l" -mod=mod -a -installsuffix cgo -o bin/backend main.go
```
diamondhands0, line 39: Note: We are choosing to do unoptimized builds here. I recommend that we keep this setting though because it allows us to super easily debug things and it didn't seem to slow things down that much. Maybe we can keep it until we have real issues.

lazynina (approved):

PR #570, opened 2024-04-01
Add a Preceding Transaction Field to Transaction Endpoints.

poolcoke:
…eProfile.


lazynina (approved):

routes/transaction.go
```diff
@@ -282,8 +284,12 @@ func (fes *APIServer) UpdateProfile(ww http.ResponseWriter, req *http.Request) {
 		_AddBadRequestError(ww, fmt.Sprintf("UpdateProfile: Problem with getUserMetadataFromGlobalState: %v", err))
 		return
 	}
-
-	utxoView, err := fes.backendServer.GetMempool().GetAugmentedUniversalView()
+	utxoView, err := lib.GetAugmentedUniversalViewWithAdditionalTransactions(
+		fes.backendServer.GetMempool(),
+		fes.backendServer.GetBlockchain().BlockTip().Height+1,
+		time.Now().UnixNano(),
```
tholonious, line 290: Same comment as in the core repo PR. I would just have the `GetAugmentedUniversalViewWithAdditionalTransactions` populate the height and timestamp internally so they're both encapsulated there. Allowing callers to provide these params will run the risk of values being used that don't reflect how the txns will connect on-chain

tholonious (approved):

routes/transaction.go
poolcoke: Addressed and fixed!

lazynina (approved):

PR #584, opened 2024-04-06
Jp/subsidized update profile

poolcoke:


routes/transaction.go
```diff
@@ -626,6 +777,238 @@ func GetBalanceForPublicKeyUsingUtxoView(
...
+}
+
+// ComputeNecessarySubsidyForTransaction is a helper function to find out how much
+// a user must be transferred in order to exactly execute a given transaction.
+//
+// In the balance model era, the total amount of DESO required to execute a transaction
+// is equivalent to the sum of the fees paid and the sum of the basic transfers (outputs).
+func ComputeNecessarySubsidyForTransaction(txn *lib.MsgDeSoTxn) uint64 {
+	necessarySubsidy := txn.TxnFeeNanos
+	for _, txOutput := range txn.TxOutputs {
```
lazynina, line 1006: We must delete this otherwise the wallet can be drained instantaneously.

routes/transaction.go
```diff
@@ -626,6 +777,238 @@ func GetBalanceForPublicKeyUsingUtxoView(
...
+		profilePublicKeyBytes,
+		requestData.NewUsername,
+		requestData.NewDescription,
+		requestData.NewProfilePic,
+		requestData.NewCreatorBasisPoints,
+		requestData.NewStakeMultipleBasisPoints,
+		requestData.IsHidden,
+		additionalFees,
+		extraData,
+		requestData.MinFeeRateNanosPerKB,
```
lazynina, line 907: We should force this txn to be constructed with the global min fee rate, otherwise the subsidizing wallet could be drained very quickly.

lazynina (changes requested):
How can we prevent abuse of the subsidy endpoint? Nefarious actors could drain the wallet very quickly.

routes/transaction.go
```diff
@@ -72,6 +73,146 @@ func (fes *APIServer) GetTxn(ww http.ResponseWriter, req *http.Request) {
...
+		TxnHashHex:               atomicTxn.Hash().String(),
+		TransactionIDBase58Check: lib.PkToString(atomicTxn.Hash()[:], fes.Params),
+	}
+
+	if err := json.NewEncoder(ww).Encode(res); err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf(
+			"SubmitAtomicTransaction: Problem encoding response as JSON: %v", err))
+		return
+	}
+}
```
lazynina, line 214: Can we move this to a separate PR such that we can accept this change independently of the subsidization changes?

routes/transaction.go
```diff
@@ -626,6 +777,238 @@ func GetBalanceForPublicKeyUsingUtxoView(
...
+			fmt.Sprintf("SubsidizedUpdateProfile: Problem generating subsidization seed: %v", err))
+		return
+	}
+	subsidizationPublicKey, subsidizationPrivateKey, _, err :=
+		lib.ComputeKeysFromSeed(subsidizationSeedBytes, 0, fes.Params)
+	if err != nil {
+		_AddBadRequestError(ww,
+			fmt.Sprintf("SubsidizedUpdateProfile: Problem computing subsidization keys: %v", err))
+		return
+	}
```
lazynina, line 830: I would move all the subsidization logic into its own function. This way we could enforce more strict checks around it and use it more broadly.

lazynina (changes requested):

routes/transaction.go
poolcoke: I wasn't sure this would be accessible to the frontend. Isn't this a risk only if we accept transactions to subsidize from the frontend?

Currently transactions are constructed AND subsidized in the backend in the same endpoint, meaning ComputeNecessarySubsidyForTransaction should be safe. Happy to change it if we want though, could be bad to have this in general.

routes/transaction.go
poolcoke: Yes! I'm going to leave this in until @diamondhands0 has a chance to read through everything. 

Something we might consider is leaving this backend PR as a draft until we decide we need it. It shouldn't be too difficult to merge this PR in later if we wanted to. It's mostly just adding new endpoints and helper functions.

routes/transaction.go
lazynina: I make a request to `subsidized-update-profile` and add `TransactionFees: [{ PublicKeyBase58Check: 'some-public-key', AmountNanos: 10*1e9}]`. Now in line 875 of your transaction.go file, this gets added as an additional output in `updateProfileTxn`. `updateProfileTxn` is passed into `ComputeNecessarySubsidyForTransaction` and I've just taken 10 DESO from the subsidy wallet.

poolcoke: Ahh yeah you're correct... didn't realize the transaction request data enabled the user to specify the transaction fees. Assumed we would compute that on the backend based on network conditions.

lazynina: yeah they aren't network fees really - they allow app developers to easily request a amount of DESO when using their app to transact.

routes/transaction.go
diamondhands0: +1

routes/transaction.go
diamondhands0: Great catch!

diamondhands0 (commented):
I reviewed this earlier and it's really great code, but I agree with @lazynina that we should keep it as a draft and merge it in later when we need it. At that time we can thoughtfully disable things like the additionalOutputs and guarantee that it's free of any issues.

I would add, though, that it's extremely valuable that you did this full end to end BE<>FE test @poolcoke because without it we would run the risk of having something critical missing in our implementation that would require another hard fork to fix.

PR #609, opened 2024-04-27
Upgrades to support DeSo AMMs and CCV2

diamondhands0:
This PR contains several SAFE upgrades to backend to support the various things we need for DeSo AMMs:
- Add a `TxnStatus` field that can be provided as an argument to `GetDAOCoinLimitOrders`, `GetTxn`, and `GetTransactorDAOCoinLimitOrders`. This param allows the caller to either consider *unconfirmed txns*, which was the existing behavior before the introduction of this field, or to only consider *confirmed* txns. The latter is what we need for the AMM since we only want to be reacting to things that are finalized. Note that this change is SAFE because the default value is TxnStatusUnconfirmed, which was the pre-existing behavior. If you don't pass the argument, as none of the existing callers do, you will get the exact same result. It's only if you pass TxnStatusConfirmed, as the AMM code does, that you will get a different result than before.
- Make it so that you can provide a list of OrderIds to `GetDAOCoinLimitOrdersRequest` and get back the ones that are currently on the book. Needed for the main loop of the AMM.
- Introduce a new function called `IsDesoPkid` that flexibly allows the caller of backend endpoints to specify any of {"DESO", MiannetZeroPkidBase58Check, TestnetZeroPkidBase58Check} and get the same behavior. This change is needed because some things rely on using a ZeroPkid while other things rely on the "DESO" string being passed. This change is SAFE because it makes these functions LESS restrictive, and all pre-existing code that calls the endpoints with "DESO" are unaffected.
- Allow for filtering by buying/selling pkid in `GetTransactorDAOCoinLimitOrders` . This is needed for the main loop of the AMM. This change is SAFE because the params are optional, and leaving them out gives you the exact same behavior as before.

### DIFF
--- a/routes/dao_coin_exchange.go
+++ b/routes/dao_coin_exchange.go
@@ -2,6 +2,7 @@ package routes
 
 import (
 	"bytes"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"github.com/deso-protocol/core/lib"
@@ -18,6 +19,13 @@ import (
 type GetDAOCoinLimitOrdersRequest struct {
 	DAOCoin1CreatorPublicKeyBase58Check string `safeForLogging:"true"`
 	DAOCoin2CreatorPublicKeyBase58Check string `safeForLogging:"true"`
+
+	// A list of hex OrderIds that we will fetch
+	OrderIds []string `safeForLogging:"true"`
+	// If unset, defaults to TxnStatusUnconfirmed. If set to "unconfirmed" we will
+	// consider all txns including those in the mempool. If set to "confirmed" then
+	// we will only consider txns that have been confirmed according to consensus.
+	TxnStatus TxnStatus `safeForLogging:"true"`
 }
 
 type GetDAOCoinLimitOrdersResponse struct {
@@ -63,8 +71,8 @@ func (fes *APIServer) GetDAOCoinLimitOrders(ww http.ResponseWriter, req *http.Re
 		return
 	}
 
-	if requestData.DAOCoin1CreatorPublicKeyBase58Check == DESOCoinIdentifierString &&
-		requestData.DAOCoin2CreatorPublicKeyBase58Check == DESOCoinIdentifierString {
+	if IsDesoPkid(requestData.DAOCoin1CreatorPublicKeyBase58Check) &&
+		IsDesoPkid(requestData.DAOCoin2CreatorPublicKeyBase58Check) {
 		_AddBadRequestError(
 			ww,
 			fmt.Sprint("GetDAOCoinLimitOrders: Must provide either a "+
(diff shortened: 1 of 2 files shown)


routes/dao_coin_exchange.go
```diff
@@ -74,16 +82,115 @@ func (fes *APIServer) GetDAOCoinLimitOrders(ww http.ResponseWriter, req *http.Re
...
+		)
+		return
+	}
+
 	utxoView, err := fes.backendServer.GetMempool().GetAugmentedUniversalView()
 	if err != nil {
 		_AddInternalServerError(ww, fmt.Sprintf("GetDAOCoinLimitOrders: Problem fetching utxoView: %v", err))
 		return
 	}
+	if txnStatus == TxnStatusConfirmed {
```
lazynina, line 104: Can we move the logic to get a utxo view to a separate helper function? Basically just take in a Boolean - something like `getUtxoView(uncommitted bool)` or maybe a different name. I would also recommend using the GetCommittedTipView function on blockchain since it's really clear what's happening, need to check if it's public tho.

routes/dao_coin_exchange.go
```diff
@@ -74,16 +82,115 @@ func (fes *APIServer) GetDAOCoinLimitOrders(ww http.ResponseWriter, req *http.Re
...
+		utxoView = lib.NewUtxoView(
+			fes.backendServer.GetBlockchain().DB(),
+			fes.Params,
+			nil,
+			fes.backendServer.GetBlockchain().Snapshot(),
+			nil)
+	}
+
+	// If they requested a specific set of OrderIds then fetch them directly
+	if len(requestData.OrderIds) > 0 {
```
lazynina, line 114: This kinda feels like we're overloading this endpoint now. Any reason we can't break it into a separate one?

routes/dao_coin_exchange.go
```diff
@@ -74,16 +82,115 @@ func (fes *APIServer) GetDAOCoinLimitOrders(ww http.ResponseWriter, req *http.Re
...
+				return
+			}
+			if orderEntry == nil {
+				continue
+			}
+			if orderEntry.IsDeleted() {
+				continue
+			}
+			orderRes, err := buildDAOCoinLimitOrderResponse(
+				lib.Base58CheckEncode(orderEntry.TransactorPKID[:], false, fes.Params),
```
lazynina, line 159: We need to look up the public keys for these PKIDs. If something had a swap identity, this would lead to unexpected behavior.

routes/dao_coin_exchange.go
```diff
@@ -74,16 +82,115 @@ func (fes *APIServer) GetDAOCoinLimitOrders(ww http.ResponseWriter, req *http.Re
 		return
 	}
 
+	txnStatus := requestData.TxnStatus
+	if txnStatus == "" {
+		txnStatus = TxnStatusUnconfirmed
+	} else if txnStatus != TxnStatusUnconfirmed &&
```
lazynina, line 88:
```suggestion
	}
   if txnStatus != TxnStatusUnconfirmed &&
```

routes/dao_coin_exchange.go
```diff
@@ -176,8 +311,45 @@ func (fes *APIServer) GetTransactorDAOCoinLimitOrders(ww http.ResponseWriter, re
 		)
 		return
 	}
+	var buyingCoinPkid *lib.PKID
+	if requestData.BuyingCoinPublicKeyBase58Check != "" {
+		if IsDesoPkid(requestData.BuyingCoinPublicKeyBase58Check) {
```
lazynina, line 316: We could probably just simplify to setting buyingCoin to zero PKID and then just have an if !IsDesoPkid condition. Also maybe we refactor this into a helper function so that the logic of getting a PKID from a public key base 58 check that may be "deso" is all in one place and easily reused everywhere.

routes/transaction.go
```diff
@@ -25,9 +25,21 @@ import (
...
 )
 
+type TxnStatus string
+
+const (
+	TxnStatusUnconfirmed TxnStatus = "unconfirmed"
+	TxnStatusConfirmed   TxnStatus = "confirmed"
+	// TODO: It would be useful to have one that is "InBlock" or something like that, which
+	// means we'll consider txns that are in unconfirmed blocks but will *not* consider txns
+	// that are in the mempool. It's a kindof middle-ground.
```
lazynina, line 35: Yeah this is a good to do and part of the reason I think creating a helper utility for getting the view makes sense such that we can introduce this functionality later. It's actually not too challenging to do an "in block" thing since we have a "get uncommitted tip view" function already on blockchain.

routes/transaction.go
```diff
@@ -25,9 +25,21 @@ import (
 	"github.com/pkg/errors"
 )
 
+type TxnStatus string
+
+const (
+	TxnStatusUnconfirmed TxnStatus = "unconfirmed"
```
lazynina, line 31: Should we change this from "unconfirmed" to "mempool"? This makes the distinction more clear if we later introduce an "in block" status which really would be better named "unconfirmed", but maybe that would be better as "unconfirmed blocks" or something.

lazynina (changes requested):

routes/dao_coin_exchange.go
```diff
@@ -18,6 +19,13 @@ import (
 type GetDAOCoinLimitOrdersRequest struct {
 	DAOCoin1CreatorPublicKeyBase58Check string `safeForLogging:"true"`
 	DAOCoin2CreatorPublicKeyBase58Check string `safeForLogging:"true"`
+
+	// A list of hex OrderIds that we will fetch
+	OrderIds []string `safeForLogging:"true"`
+	// If unset, defaults to TxnStatusUnconfirmed. If set to "unconfirmed" we will
+	// consider all txns including those in the mempool. If set to "confirmed" then
+	// we will only consider txns that have been confirmed according to consensus.
```
tholonious, line 27:
```suggestion
	// we will only consider txns that have been confirmed according on the blockchain.
```

routes/transaction.go
tholonious: +1 to this. We have three statuses in the happy path for the life cycle of a txn: "InMempool", "Unconfirmed", "Committed". 

It's good to disambiguate them now.

routes/dao_coin_exchange.go
```diff
@@ -18,6 +19,13 @@ import (
 type GetDAOCoinLimitOrdersRequest struct {
 	DAOCoin1CreatorPublicKeyBase58Check string `safeForLogging:"true"`
 	DAOCoin2CreatorPublicKeyBase58Check string `safeForLogging:"true"`
+
+	// A list of hex OrderIds that we will fetch
+	OrderIds []string `safeForLogging:"true"`
```
tholonious, line 24: +1 to Nina's comment. Please create a new endpoint that takes in `OrderIds` and returns the exact orders. Overloading the existing endpoint with more orthogonal fields will make its usage more error prone.

lazynina (approved):

lazynina:
### Merge activity

* **Apr 29, 2:10 PM EDT**: @lazynina started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/deso-protocol/backend/609).

PR #628, opened 2024-05-23
Add simple balance endpoint (for amm usage)

diamondhands0:
There wasn't a clean way to get the balance of a set of coins, which is needed in order to determine if the amm should initialize a market. Added this endpoint to address this need.

There's also a thorough test for it in the amm code here:
https://github.com/deso-protocol/amm-service/blob/f4953a8f1a141110a10f2ced44e7983176d5d46e/ammlib/amm_config_test.go#L1170

### DIFF
--- a/routes/server.go
+++ b/routes/server.go
@@ -2427,7 +2427,7 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
 func CheckPrecedingTransactions(inner http.Handler, precedingTransactionsLimit int) http.Handler {
 	return http.HandlerFunc(func(ww http.ResponseWriter, rr *http.Request) {
 		// If the request is NOT a POST request, skip this middleware as the request has no relevant payload.
-		if rr.Method != "POST" {
+		if rr.Method != "POST" || rr.Header.Get("Content-Type") != "application/json" {
 			inner.ServeHTTP(ww, rr)
 			return
 		}
--- a/routes/server.go
+++ b/routes/server.go
@@ -77,7 +77,7 @@ const (
 	RoutePathGetSingleProfile                           = "/api/v0/get-single-profile"
 	RoutePathGetSingleProfilePicture                    = "/api/v0/get-single-profile-picture"
 	RoutePathGetHodlersForPublicKey                     = "/api/v0/get-hodlers-for-public-key"
-	RoutePathGetBalancesForPublicKey                    = "/api/v0/get-balances-for-public-key"
+	RoutePathGetTokenBalancesForPublicKey               = "/api/v0/get-token-balances-for-public-key"
 	RoutePathGetHodlersCountForPublicKeys               = "/api/v0/get-hodlers-count-for-public-keys"
 	RoutePathGetDiamondsForPublicKey                    = "/api/v0/get-diamonds-for-public-key"
 	RoutePathGetFollowsStateless                        = "/api/v0/get-follows-stateless"
@@ -931,10 +931,10 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
 			PublicAccess,
 		},
 		{
-			"GetBalancesForPublicKey",
+			"GetTokenBalancesForPublicKey",
 			[]string{"POST", "OPTIONS"},
-			RoutePathGetBalancesForPublicKey,
-			fes.GetBalancesForPublicKey,
+			RoutePathGetTokenBalancesForPublicKey,
+			fes.GetTokenBalancesForPublicKey,
 			PublicAccess,
 		},
 		{
--- a/routes/server.go
+++ b/routes/server.go
@@ -77,6 +77,7 @@ const (
 	RoutePathGetSingleProfile                           = "/api/v0/get-single-profile"
 	RoutePathGetSingleProfilePicture                    = "/api/v0/get-single-profile-picture"
 	RoutePathGetHodlersForPublicKey                     = "/api/v0/get-hodlers-for-public-key"
+	RoutePathGetBalancesForPublicKey                    = "/api/v0/get-balances-for-public-key"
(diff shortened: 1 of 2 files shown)


routes/server.go
```diff
@@ -77,6 +77,7 @@ const (
 	RoutePathGetSingleProfile                           = "/api/v0/get-single-profile"
 	RoutePathGetSingleProfilePicture                    = "/api/v0/get-single-profile-picture"
 	RoutePathGetHodlersForPublicKey                     = "/api/v0/get-hodlers-for-public-key"
+	RoutePathGetBalancesForPublicKey                    = "/api/v0/get-balances-for-public-key"
```
lazynina, line 80:
```suggestion
	RoutePathGetTokenBalancesForPublicKey                    = "/api/v0/get-token-balances-for-public-key"
```
Can we make it more clear that this is for DeSo Token (FKA DAO coin) balances?

routes/server.go
```diff
@@ -929,6 +930,13 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
 			fes.GetHodlersForPublicKey,
 			PublicAccess,
 		},
+		{
+			"GetBalancesForPublicKey",
+			[]string{"POST", "OPTIONS"},
+			RoutePathGetBalancesForPublicKey,
+			fes.GetBalancesForPublicKey,
```
lazynina, line 937:
```suggestion
			"GetTokenBalancesForPublicKey",
			[]string{"POST", "OPTIONS"},
			RoutePathGetTokenBalancesForPublicKey,
			fes.GetTokenBalancesForPublicKey,
```

routes/user.go
```diff
@@ -1479,6 +1479,119 @@ func (fes *APIServer) GetHodlersForPublicKey(ww http.ResponseWriter, req *http.R
...
+	// A base-10 encoded string of the balance in base units, which is internally
+	// represented as a uint256
+	BalanceBaseUnits string
+}
+
+type GetBalancesForPublicKeyResponse struct {
+	Balances map[string]*SimpleBalanceResponse
+}
+
+func (fes *APIServer) GetBalancesForPublicKey(ww http.ResponseWriter, req *http.Request) {
```
lazynina, line 1511:
```suggestion
func (fes *APIServer) GetTokenBalancesForPublicKey(ww http.ResponseWriter, req *http.Request) {
```

routes/user.go
```diff
@@ -1479,6 +1479,119 @@ func (fes *APIServer) GetHodlersForPublicKey(ww http.ResponseWriter, req *http.R
...
+	}
+
+	// Decode the public key for which we are fetching balances
+	if requestData.UserPublicKey == "" {
+		_AddBadRequestError(ww, fmt.Sprintf("GetBalancesForPublicKey: Missing UserPublicKey"))
+		return
+	}
+	userPublicKeyBytes, _, err := lib.Base58CheckDecode(requestData.UserPublicKey)
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("GetHodlersForPublicKey: Problem decoding user public key: %v", err))
```
lazynina, line 1538: Clean up error prefixes

routes/user.go
```diff
@@ -1479,6 +1479,119 @@ func (fes *APIServer) GetHodlersForPublicKey(ww http.ResponseWriter, req *http.R
...
+	// the balance of for the user. For DESO,
+	// can specify either "DESO" or the ZeroPkid via base58check encoding.
+	// Either will work.
+	CreatorPublicKeys []string
+
+	// If unset, defaults to TxnStatusInMempool
+	TxnStatus TxnStatus `safeForLogging:"true"`
+}
+
+type SimpleBalanceResponse struct {
```
lazynina, line 1496:
```suggestion
type SimpleTokenBalanceResponse struct {
```

routes/user.go
```diff
@@ -1479,6 +1479,119 @@ func (fes *APIServer) GetHodlersForPublicKey(ww http.ResponseWriter, req *http.R
...
+		balanceEntry, _, _ := utxoView.GetBalanceEntryForHODLerPubKeyAndCreatorPubKey(
+			userPublicKeyBytes, creatorPkBytes, true)
+		if balanceEntry == nil || balanceEntry.IsDeleted() {
+			balanceEntry = &lib.BalanceEntry{}
+		}
+		// Convert balanceEntry uint256 to string
+		balancesMap[creatorPublicKeyStr] = &SimpleBalanceResponse{
+			UserPublicKeyBase58Check:    requestData.UserPublicKey,
+			CreatorPublicKeyBase58Check: creatorPublicKeyStr,
+			BalanceBaseUnits:            balanceEntry.BalanceNanos.String(),
```
lazynina, line 1581: Would returning the hex string here also be useful? So we could do `BalanceBaseUnits` and `BalanceBaseUnitsHex` or something. Not sure if you're concerned about losing precision when going to base10 string.

lazynina (changes requested):

routes/server.go
diamondhands0: Fixed.

routes/server.go
diamondhands0: Fixed.

routes/user.go
diamondhands0: Fixed.

routes/user.go
diamondhands0: Fixed.

routes/user.go
diamondhands0: Fixed.

routes/user.go
diamondhands0: We don't lose precision when we convert to a base-10 string. It's the same value, just easier to read.

lazynina (approved):

PR #656, opened 2024-07-24
include inner txn hexes in atomic txn construction endpoints

lazynina:


routes/atomic_txns.go
```diff
@@ -92,12 +96,23 @@ func (fes *APIServer) CreateAtomicTxnsWrapper(ww http.ResponseWriter, req *http.
 		}
 	}
 
+	innerTransactionHexes := []string{}
```
diamondhands0, line 99: Since we do this in a few places, can we just break it out into a function like `GetInnerTarnsactionHexes(txn *MsgDeSoTxn)` and call it in all three places?

diamondhands0 (approved):

PR #669, opened 2024-08-31
Updates to support smoother AMM coin creation in Focus

diamondhands0:
- Add AmmMetadataPublicKey as an optional backend flag. This flag is required to be set if you want to SET or GET trading fees for DaoCoinLimitOrderWithFee.
- Updated trading fees to be an association set by the AmmMetadataPublicKey rather than by the individual user. This allows users to credibly disable trading fee updates. Tests are in the amm-service repo and all are passing.
- Fixed a bunch of issues related to using DESO as the quote currency in a DaoCoinLimitOrderWithFees transaction. Thorough and updated tests are in the amm-service repo and all tests are passing there.
- Added a new backend endpoint to get all holders of a coin WITH locked balances as well. This was needed in order to support lockup-based features when creating a coin, without having to add a dependency on GraphQL (because the amm-service currenly has no deps on GraphQL).

### DIFF
--- a/routes/dao_coin_exchange_with_fees.go
+++ b/routes/dao_coin_exchange_with_fees.go
@@ -501,11 +501,11 @@ type GetBaseCurrencyPriceResponse struct {
 	BestBidInUsd            float64 `safeForLogging:"true"`
 
 	// Useful for computing "cashout" values on the wallet page
-	ExecutionAmountInBaseCurrency float64 `safeForLogging:"true"`
-	ReceiveAmountInQuoteCurrency  float64 `safeForLogging:"true"`
-	ReceiveAmountInUsd            float64 `safeForLogging:"true"`
-	ExecutionPriceInQuoteCurrency float64 `safeForLogging:"true"`
-	ExecutionPriceInUsd           float64 `safeForLogging:"true"`
+	ExecutionAmountInBaseCurrency string `safeForLogging:"true"`
+	ReceiveAmountInQuoteCurrency  string `safeForLogging:"true"`
+	ReceiveAmountInUsd            string `safeForLogging:"true"`
+	ExecutionPriceInQuoteCurrency string `safeForLogging:"true"`
+	ExecutionPriceInUsd           string `safeForLogging:"true"`
 }
 
 func (fes *APIServer) GetBaseCurrencyPriceEndpoint(ww http.ResponseWriter, req *http.Request) {
@@ -681,28 +681,35 @@ func (fes *APIServer) GetBaseCurrencyPriceEndpoint(ww http.ResponseWriter, req *
 
 	// Iterate through the bids "filling" orders until we hit the base currency
 	// quantity we're looking for.
-	baseCurrencyFilled := float64(0.0)
-	baseCurrencyToFill := requestData.BaseCurrencyQuantityToSell
-	quoteCurrencyReceived := float64(0.0)
+	baseCurrencyFilled := big.NewFloat(0.0)
+	baseCurrencyToFill := big.NewFloat(requestData.BaseCurrencyQuantityToSell)
+	quoteCurrencyReceived := big.NewFloat(0.0)
 	for _, bid := range simpleBidOrders {
 		// If the amount filled plus the amount we're about to fill is greater
 		// than the amount we're looking to fill, then we just partially fill
 		// the order.
-		if baseCurrencyFilled+bid.AmountInBaseCurrency > baseCurrencyToFill {
-			baseCurrencyFromThisOrder := baseCurrencyToFill - baseCurrencyFilled
-			quoteCurrencyReceived += baseCurrencyFromThisOrder * bid.PriceInQuoteCurrency
+		if big.NewFloat(0.0).Add(
+			baseCurrencyFilled,
+			big.NewFloat(bid.AmountInBaseCurrency),
+		).Cmp(baseCurrencyToFill) > 0 {
+			baseCurrencyFromThisOrder := big.NewFloat(0).Sub(baseCurrencyToFill, baseCurrencyFilled)
+			addlQuoteCurrencyReceivedFromOrdered := big.NewFloat(0).Mul(baseCurrencyFromThisOrder, big.NewFloat(bid.PriceInQuoteCurrency))
+			quoteCurrencyReceived = big.NewFloat(0).Add(quoteCurrencyReceived, addlQuoteCurrencyReceivedFromOrdered)
+			// Since we've fully filled the order, we can simply set the base currency filled
+			// to the amount we're looking to fill from the request.
 			baseCurrencyFilled = baseCurrencyToFill
 			break
 		}
-		quoteCurrencyReceived += bid.AmountInBaseCurrency * bid.PriceInQuoteCurrency
-		baseCurrencyFilled += bid.AmountInBaseCurrency
+		addlQuoteCurrencyReceivedFromOrder := big.NewFloat(0).Mul(big.NewFloat(bid.AmountInBaseCurrency), big.NewFloat(bid.PriceInQuoteCurrency))
+		quoteCurrencyReceived = big.NewFloat(0).Add(quoteCurrencyReceived, addlQuoteCurrencyReceivedFromOrder)
+		baseCurrencyFilled = big.NewFloat(0).Add(baseCurrencyFilled, big.NewFloat(bid.AmountInBaseCurrency))
 	}
 
 	// Now the amount to fill and the quote currency received should be ready to go
 	// so we can compute the price.
-	priceInQuoteCurrency := 0.0
-	if baseCurrencyFilled > 0 {
-		priceInQuoteCurrency = quoteCurrencyReceived / baseCurrencyFilled
+	priceInQuoteCurrency := big.NewFloat(0.0)
+	if baseCurrencyFilled.Sign() > 0 {
+		priceInQuoteCurrency = big.NewFloat(0).Quo(quoteCurrencyReceived, baseCurrencyFilled)
 	}
 
 	// Get the price of the quote currency in usd. Use the mid price
@@ -730,11 +737,11 @@ func (fes *APIServer) GetBaseCurrencyPriceEndpoint(ww http.ResponseWriter, req *
 		BestBidInUsd:            bestBidPriceInQuoteCurrency * quoteCurrencyPriceInUsd,
 
 		// Useful for computing "cashout" values on the wallet page
-		ExecutionAmountInBaseCurrency: baseCurrencyFilled,
-		ReceiveAmountInQuoteCurrency:  quoteCurrencyReceived,
-		ReceiveAmountInUsd:            quoteCurrencyReceived * quoteCurrencyPriceInUsd,
-		ExecutionPriceInQuoteCurrency: priceInQuoteCurrency,
-		ExecutionPriceInUsd:           priceInQuoteCurrency * quoteCurrencyPriceInUsd,
+		ExecutionAmountInBaseCurrency: baseCurrencyFilled.String(),
+		ReceiveAmountInQuoteCurrency:  quoteCurrencyReceived.String(),
+		ReceiveAmountInUsd:            big.NewFloat(0).Mul(quoteCurrencyReceived, big.NewFloat(quoteCurrencyPriceInUsd)).String(),
+		ExecutionPriceInQuoteCurrency: priceInQuoteCurrency.String(),
+		ExecutionPriceInUsd:           big.NewFloat(0).Mul(priceInQuoteCurrency, big.NewFloat(quoteCurrencyPriceInUsd)).String(),
 	}
 	if err := json.NewEncoder(ww).Encode(res); err != nil {
 		_AddBadRequestError(ww, fmt.Sprintf("GetQuoteCurrencyPriceInUsd: Problem encoding response: %v", err))
--- a/routes/dao_coin_exchange_with_fees.go
+++ b/routes/dao_coin_exchange_with_fees.go
@@ -259,7 +259,8 @@ func (fes *APIServer) UpdateDaoCoinMarketFees(ww http.ResponseWriter, req *http.
 		additionalOutputs,
 	)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


routes/dao_coin_exchange_with_fees.go
```diff
@@ -74,23 +72,47 @@ func ValidateTradingFeeMap(feeMap map[string]uint64) error {
 	return nil
 }
 
+func IsExtraDataValueTrueString(extraData map[string]string, key string) bool {
+	newMap := make(map[string][]byte)
+	for k, v := range extraData {
+		newMap[k] = []byte(v)
+	}
```
lazynina, line 79: It seems a little unnecessary to make a copy of the entire map. If there are a lot of KVs in the extra data, this may be slow. Can we just make a copy of the value we need?

routes/dao_coin_exchange_with_fees.go
```diff
@@ -74,23 +72,47 @@ func ValidateTradingFeeMap(feeMap map[string]uint64) error {
...
+func SetExtraDataValueTrue(extraData map[string][]byte, key string) {
+	extraData[key] = []byte{1}
+}
+
+func SetExtraDataValueTrueString(extraData map[string]string, key string) {
+	extraData[key] = string([]byte{1})
+}
+
 func (fes *APIServer) UpdateDaoCoinMarketFees(ww http.ResponseWriter, req *http.Request) {
+	if fes.Config.AmmMetadataPublicKey == "" {
```
lazynina, line 103: If there a reason we want this configured at the node level? Could the amm's public key be specified in the request body?

routes/dao_coin_exchange_with_fees.go
```diff
@@ -178,33 +196,70 @@ func (fes *APIServer) UpdateDaoCoinMarketFees(ww http.ResponseWriter, req *http.
...
+		_AddBadRequestError(ww, fmt.Sprintf(
+			"UpdateDaoCoinMarketFees: Problem decoding public key %s: %v",
+			fes.Config.AmmMetadataPublicKey, err))
+		return
+	}
+
+	// Fetch the existing fee association to check if fee updates are still allowed.
+	associationFilter := &lib.UserAssociationQuery{
+		AssociationType:  []byte(lib.TokenTradingFeesByPkidMapKey),
+		AssociationValue: []byte(lib.TokenTradingFeesByPkidMapKey),
```
lazynina, line 216: why are the type and value the same? I guess it's fine, just a little odd.

routes/dao_coin_exchange_with_fees.go
```diff
@@ -178,33 +196,70 @@ func (fes *APIServer) UpdateDaoCoinMarketFees(ww http.ResponseWriter, req *http.
...
+		if IsExtraDataValueTrue(association.ExtraData, lib.DisableTradingFeeUpdateKey) {
+			_AddBadRequestError(ww, fmt.Sprintf(
+				"UpdateDaoCoinMarketFees: Trading fee updates are disabled for this market"))
+			return
+		}
+	} else if len(associations) != 0 {
+		_AddBadRequestError(ww, fmt.Sprintf(
+			"UpdateDaoCoinMarketFees: Expected at most one association but got %d", len(associations)))
+		return
+	}
```
lazynina, line 237: Can we write it like:
1. `if len(associations) > 1`, throw error
2. `if len(associations) == 1`, check extra data value true thing.

I prefer to get rid of else ifs if we can.

routes/dao_coin_exchange_with_fees.go
```diff
@@ -178,33 +196,70 @@ func (fes *APIServer) UpdateDaoCoinMarketFees(ww http.ResponseWriter, req *http.
...
+			AssociationValue:    []byte(lib.TokenTradingFeesByPkidMapKey),
+		},
+		additionalExtraData,
+		requestData.MinFeeRateNanosPerKB,
+		fes.backendServer.GetMempool(),
+		additionalOutputs,
+	)
 	if err != nil {
-		_AddBadRequestError(ww, fmt.Sprintf("UpdateDaoCoinMarketFees: Problem creating transaction: %v", err))
+		_AddInternalServerError(ww, fmt.Sprintf("CreateUserAssociation: problem creating txn: %v", err))
```
lazynina, line 262: Let's make this error more clear. I think it should be `UpdateDaoCoinMarketFees: problem creating user association txn: %v`

routes/dao_coin_exchange_with_fees.go
```diff
@@ -725,6 +827,13 @@ func (fes *APIServer) GetBaseCurrencyPriceEndpoint(ww http.ResponseWriter, req *
 		return
 	}
 
+	baseCurrencyFilledFloat, _ := baseCurrencyFilled.Float64()
+	quoteCurrencyReceivedFloat, _ := quoteCurrencyReceived.Float64()
+	receiveAmountInUsdFloat, _ := big.NewFloat(0).Mul(
+		quoteCurrencyReceived, big.NewFloat(quoteCurrencyPriceInUsd)).Float64()
+	priceInQuoteCurrencyFloat, _ := priceInQuoteCurrency.Float64()
```
lazynina, line 834: what if any of these are too big for float64?

routes/transaction.go
```diff
@@ -89,7 +89,7 @@ func (fes *APIServer) GetTxn(ww http.ResponseWriter, req *http.Request) {
 	// committed tip height. Otherwise we'll be missing ~2 blocks in limbo.
 	coreChainTipHeight := fes.TXIndex.CoreChain.BlockTip().Height
 	for fes.TXIndex.TXIndexChain.BlockTip().Height < coreChainTipHeight {
-		if time.Since(startTime) > 30*time.Second {
+		if time.Since(startTime) > 90*time.Second {
```
lazynina, line 92: Please revert to 30s. I don't want a bunch of requests hanging around for 90s - it could be bad for the node.

routes/server.go
```diff
@@ -950,6 +951,13 @@ func (fes *APIServer) NewRouter() *muxtrace.Router {
 			fes.GetHodlersForPublicKey,
 			PublicAccess,
 		},
+		{
+			"GetHoldersForPublicKeyWithLockedBalances",
+			[]string{"POST", "OPTIONS"},
```
lazynina, line 956: From the looks of it, this could be a "GET" request. Any reason why you want a POST instead?

routes/user.go
```diff
@@ -410,7 +410,8 @@ func (fes *APIServer) GetHodlingsForPublicKey(
 func (fes *APIServer) GetHodlYouMap(pkid *lib.PKIDEntry, fetchProfiles bool, isDAOCoin bool, utxoView *lib.UtxoView) (
 	_youHodlMap map[string]*BalanceEntryResponse, _err error) {
 	// Get all the hodlings for this user from the db
-	entriesHodlingYou, profileHodlingYou, err := utxoView.GetHolders(pkid.PKID, fetchProfiles, isDAOCoin)
+	entriesHodlingYou, profileHodlingYou, _, _, err := utxoView.GetHolders(
```
lazynina, line 413: See comments in [here](https://github.com/deso-protocol/core/pull/1404#pullrequestreview-2313706611) about GetHolders

routes/user.go
```diff
@@ -3775,3 +3776,134 @@ func (fes *APIServer) GetProfileEntryResponseForPublicKeyBytes(publicKeyBytes []
...
+	// Get a view
+	utxoView, err := fes.backendServer.GetMempool().GetAugmentedUniversalView()
+	if err != nil {
+		_AddBadRequestError(ww, fmt.Sprintf("GetHoldersForPublicKeyWithLockedBalances: Error getting utxoView: %v", err))
+		return
+	}
+
+	// Decode the public key for which we are fetching hodlers / hodlings.  If public key is not provided, use username
+	var creatorPublicKeyBytes []byte
+	if requestData.PublicKeyBase58Check != "" {
```
lazynina, line 3810: You may be able to replace this `if else` with `GetPubKeyAndProfileEntryForUsernameOrPublicKeyBase58Check`

routes/user.go
```diff
@@ -3775,3 +3776,134 @@ func (fes *APIServer) GetProfileEntryResponseForPublicKeyBytes(publicKeyBytes []
...
+
+		// Return an error if we failed to find a profile entry
+		if profileEntry == nil {
+			_AddNotFoundError(ww, fmt.Sprintf("GetHoldersForPublicKeyWithLockedBalances: could not find profile for username: %v", username))
+			return
+		}
+		creatorPublicKeyBytes = profileEntry.PublicKey
+	}
+
+	unlockedBalanceEntrys, _, lockedBalanceEntryMap, _, err := utxoView.GetHolders(
```
lazynina, line 3828: If this is the only place where we need both the unlocked and locked balance entries, I strongly recommend we break out getting the locked balance entries in core into its own function.

routes/user.go
```diff
@@ -3775,3 +3776,134 @@ func (fes *APIServer) GetProfileEntryResponseForPublicKeyBytes(publicKeyBytes []
...
+			totalBalanceBaseUnits = uint256.NewInt().Add(
+				totalBalanceBaseUnits, &lockedBalanceEntry.BalanceBaseUnits)
+		}
+
+		if prevBalance == nil {
+			// If there's no unlocked balance for this key, we need to create the
+			// entry from scratch.
+			unlockedBalanceEntryResponse := fes._balanceEntryToResponse(
+				&lib.BalanceEntry{
+					HODLerPKID:   &pkid,
```
lazynina, line 3874: should we use `.NewPKID()` to create a copy here?

routes/dao_coin_exchange_with_fees.go
```diff
@@ -112,28 +134,24 @@ func (fes *APIServer) UpdateDaoCoinMarketFees(ww http.ResponseWriter, req *http.
...
+			"UpdateDaoCoinMarketFees: ProfilePublicKeyBase58Check and UpdaterPublicKeyBase58Check must be provided"))
+		return
 	}
 
-	// Get the public key.
-	profilePublicKey := updaterPublicKeyBytes
-	if requestData.ProfilePublicKeyBase58Check != "" {
-		profilePublicKey = profilePublicKeyBytes
+	// Decode the profile public key.
+	profilePublicKeyBytes, _, err := lib.Base58CheckDecode(requestData.ProfilePublicKeyBase58Check)
```
lazynina, line 144: You could remove some duplicated code here by using `GetPubKeyBytesFromBase58Check`

routes/dao_coin_exchange_with_fees.go
```diff
@@ -178,33 +196,70 @@ func (fes *APIServer) UpdateDaoCoinMarketFees(ww http.ResponseWriter, req *http.
 	additionalExtraData := make(map[string][]byte)
 	additionalExtraData[lib.TokenTradingFeesByPkidMapKey] = feeMapByPkidBytes
 
+	// Setting this byte makes it impossible to update the trading fee map in the
+	// future. Use with caution.
+	if requestData.DisableTradingFeeUpdate {
+		SetExtraDataValueTrue(additionalExtraData, lib.DisableTradingFeeUpdateKey)
+	}
+
+	ammPublicKeyBytes, _, err := lib.Base58CheckDecode(fes.Config.AmmMetadataPublicKey)
```
lazynina, line 205: Maybe consider using `GetPubKeyBytesFromBase58Check`

routes/dao_coin_exchange_with_fees.go
```diff
@@ -235,47 +290,86 @@ type GetDaoCoinMarketFeesRequest struct {
...
-	if existingProfileEntry == nil || existingProfileEntry.IsDeleted() {
-		return nil, fmt.Errorf(
+	profilePkid := utxoView.GetPKIDForPublicKey(profilePublicKeyBytes)
+	if profilePkid == nil {
+		return nil, false, fmt.Errorf(
 			"GetTradingFeesForMarket: Profile for public key %v does not exist",
 			profilePublicKey)
 	}
 
+	ammMetadataPublicKeyBytes, _, err := lib.Base58CheckDecode(ammMetadataPublicKey)
```
lazynina, line 320: `GetPubKeyBytesFromBase58Check`

lazynina (changes requested):

routes/server.go
lazynina: It prevents us from getting the benefits of caching. Cloudflare can help with caching GET requests.

routes/dao_coin_exchange_with_fees.go
```diff
@@ -934,6 +1051,86 @@ func InvertPriceStr(priceStr string) (string, error) {
...
+	}
+	ammPkid := utxoView.GetPKIDForPublicKey(ammPubkeyBytes)
+
+	associationQuery := &lib.UserAssociationQuery{
+		TransactorPKID:   transactorPkid.PKID,
+		TargetUserPKID:   coinPkid.PKID,
+		AppPKID:          ammPkid.PKID,
+		AssociationType:  []byte(lib.DeSoTokenWhitelistAssociationKey),
+		AssociationValue: []byte(lib.DeSoTokenWhitelistAssociationKey),
+		Limit:            10,
```
lazynina, line 1099: can we make this limit = 1?

routes/dao_coin_exchange_with_fees.go
```diff
@@ -1214,6 +1412,17 @@ func (fes *APIServer) HandleMarketOrder(
 		return nil, fmt.Errorf("HandleMarketOrder: Problem decoding public key %s: %v",
 			req.TransactorPublicKeyBase58Check, err)
 	}
+	// Create a transaction to whitelist the token in the user's wallet if it's not already,
+	// and if that's desired.
+	tokenWhitelistTxn, err := fes.MaybeCreateTokenWhitelistAssociation(
```
lazynina, line 1417: can we wrap this in a `if !skipWhitelist`? I think we don't use this txn if skipWhitelist is true.

routes/user.go
lazynina: It replaces the `else` case as well! The other reason I recommend using these is in the event that we want to broadly add more validations. reducing code duplication makes my life easier when we have to introduce new constraints.

lazynina (approved):

PR #670, opened 2024-09-10
Fix hot feed

diamondhands0:
There was an issue preventing posts older than 4h from being considered for the hot feed, caused by the shift from 5m per block to 1.5s per block (it was originally 60d then it became 4h).

### DIFF
--- a/routes/hot_feed.go
+++ b/routes/hot_feed.go
@@ -24,15 +24,18 @@ import (
 // HotnessFeed scoring algorithm knobs.
 const (
 	// Number of blocks per halving for the scoring time decay for the global hot feed.
-	DefaultHotFeedTimeDecayBlocks uint64 = 72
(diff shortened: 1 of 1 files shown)


routes/hot_feed.go
```diff
@@ -555,9 +557,30 @@ func (fes *APIServer) PopulateHotnessInfoMap(
 			// to posts from the specified look-back period without even looking up the post time stamp.
 			isCreatePost, postHashCreated := CheckTxnForCreatePost(txn)
 			if isCreatePost {
+				// We start with the creator's balance in computing the score
```
diamondhands0, line 560: This part isn't that important, it just makes the creator's balance factor into a post's score, which offers a slight improvement.

superzordon (approved):

superzordon (approved):

PR #694, opened 2024-10-30
Ln/update quote currency endpoint for deso

lazynina:


routes/dao_coin_exchange_with_fees.go
```diff
@@ -908,22 +908,31 @@ const FOCUS_FLOOR_PRICE_DESO_NANOS = 166666
...
-		price := fmt.Sprintf("%0.9f", float64(desoUsdCents)/100)
-		return price, price, price, nil // TODO: get real bid and ask prices.
-	}
 	utxoView, err := fes.backendServer.GetMempool().GetAugmentedUniversalView()
 	if err != nil {
 		return "", "", "", fmt.Errorf(
 			"GetQuoteCurrencyPriceInUsd: Error fetching mempool view: %v", err)
 	}
+	if IsDesoPkid(quoteCurrencyPublicKey) {
+		usdcProfileEntry := utxoView.GetProfileEntryForUsername([]byte("dusdc_"))
```
diamondhands0, line 917: Oof. Can we please pull this out into a variable somewhere at least? It's referenced below in this function as well I think.

diamondhands0 (approved):
Love this change. Makes everything consistent.

routes/dao_coin_exchange_with_fees.go
lazynina: made a const called `dusdcProfileUsername` for this.

PR #708, opened 2024-12-17
Generalize GetBaseCurrencyPriceEndpoint to take multiple entries

diamondhands0:


lazynina (approved):
I think it's fine. If we hit performance issues, we'll address them then.

PR #710, opened 2025-01-15
Add ability to submit signatures without embedding them into txns

diamondhands0:
I'm updating the Openfund docs and working on a tutorial to enable algorithmic trading, and this change makes it possible to trivially sign transactions in languages that we don't support, such as Python, which I was using for the tutorial (most commonly used by traders).

This code is now fully tested by the amm-service, which I adapted to use the alternative pathways in a couple of tests. See this PR here:
https://github.com/deso-protocol/amm-service/pull/15


lazynina (approved):
Approved w/ some minor nits on error messages.

PR #720, opened 2025-03-12
Add get txns endpoint.

mattfoley8:


routes/transaction.go
```diff
@@ -122,6 +122,116 @@ func (fes *APIServer) GetTxn(ww http.ResponseWriter, req *http.Request) {
...
+	// Decode the TxnHashHexes.
+	var txnHashes []*lib.BlockHash
+	for _, txnHashHex := range requestData.TxnHashHexes {
+		txnHashBytes, err := hex.DecodeString(txnHashHex)
+		if err != nil || len(txnHashBytes) != lib.HashSizeBytes {
+			_AddBadRequestError(ww, fmt.Sprintf("GetTxns: Error parsing txn hash %s: %v", txnHashHex, err))
+			return
+		}
+		txnHashes = append(txnHashes, lib.NewBlockHash(txnHashBytes))
+	}
```
lazynina, line 181:
```suggestion
	txnHashes := make(map[string]*lib.BlockHash, len(requestData.TxnHashHexes))
	for _, txnHashHex := range requestData.TxnHashHexes {
		txnHashBytes, err := hex.DecodeString(txnHashHex)
		if err != nil || len(txnHashBytes) != lib.HashSizeBytes {
			_AddBadRequestError(ww, fmt.Sprintf("GetTxns: Error parsing txn hash %s: %v", txnHashHex, err))
			return
		}
		txnHashes[txnHashHex] = lib.NewBlockHash(txnHashBytes)
	}
```
This helps with two things: 
1. If we have duplicates, we only really need to go to the DB onces.
2. the `.String` function on the `BlockHash` type isn't SUPER fast, so we can avoid calling it below since we already have the txn hash hex for it.

routes/transaction.go
```diff
@@ -122,6 +122,116 @@ func (fes *APIServer) GetTxn(ww http.ResponseWriter, req *http.Request) {
...
+	// endpoint to incorrectly report that the txn doesn't exist on the node, when in
+	// fact it is in "limbo" between the mempool and txindex.
+	//
+	// 1. Check the mempool for each txn if TxnStatusInMempool.
+	txnsFoundInMempool := make(map[string]bool)
+	if txnStatus == TxnStatusInMempool {
+		mempool := fes.backendServer.GetMempool()
+		for _, txnHash := range txnHashes {
+			txnsFoundInMempool[txnHash.String()] = mempool.IsTransactionInPool(txnHash)
+		}
```
lazynina, line 200:
```suggestion
		for txHashHex, txnHash := range txnHashes {
			txnsFoundInMempool[txHashHex] = mempool.IsTransactionInPool(txnHash)
		}
```

routes/transaction.go
```diff
@@ -122,6 +122,116 @@ func (fes *APIServer) GetTxn(ww http.ResponseWriter, req *http.Request) {
...
+	for _, txnHash := range txnHashes {
+		txnsFoundInTxindex[txnHash.String()] = lib.DbCheckTxnExistence(fes.TXIndex.TXIndexChain.DB(), nil, txnHash)
+	}
+
+	// Construct response.
+	res := &GetTxnsResponse{TxnsFound: make(map[string]bool)}
+	for _, txnHash := range txnHashes {
+		// If TxnStatusCommited, txnsFoundInMempool[txnHash.String()] will always be false.
+		res.TxnsFound[txnHash.String()] = txnsFoundInMempool[txnHash.String()] || txnsFoundInTxindex[txnHash.String()]
+	}
```
lazynina, line 226: if `TxnStatus` is `Committed`, then we can simply return `txnsFoundInTxindex`. We only need to do this merge if `TxnStatus === InMempool`. And if we need to do the merge, then we can use `TxnHashHex` instead of `.String()`.

lazynina (changes requested):

lazynina (approved):

